PR #5, opened 2026-05-01
feat(schema): POST /schema command endpoint

shoriminimoe:
## Summary

Implements the `POST /schema` command endpoint per `docs/superpowers/specs/2026-05-01-schema-endpoint-design.md` (ADR-008 / ADR-013).

- **18 commands** (4 entity kinds × 4–5 verbs) routed via a discriminated msgspec union (`SchemaRequest`), dispatched by a `dict[type, Handler]` table — no 17-arm `match`.
- **Single-transaction contract**: each handler reads the projection, validates the state transition, mutates the projection, and appends a `schema_change_log` row. Litestar's `before_send_handler="autocommit"` commits on 2xx and rolls back on 4xx (verified by an end-to-end test that asserts `schema_version` does not advance through a failed command).
- **Stable error envelope** with three categories (400 `invalid_request`, 404 `not_found`, 409 `conflict`) and a typed `ErrorCode` enum (`payload_no_changes`, `definition_required`, `name_reserved`, `name_held_by_tombstone`, `use_update`, `parent_type_not_found`, `entity_not_found`, `missing_required_field`, `invalid_payload_shape`).
- **OpenAPI**: every command and its payload shape are published as a `oneOf` discriminator on the request body; the `SchemaErrorResponse` envelope is attached to 400/404/409 responses. A test asserts all 18 command tags appear.
- **Validation matrix** from ADR-008 enumerated cell-by-cell across 4 handler test files (~50 handler tests).

## Notable design points

- **Layering**: per the project's hexagonal convention, `db/` imports `advanced_alchemy` core only; the `domain/schema/services/` and `domain/schema/controllers/` layer reaches into `advanced_alchemy.extensions.litestar`.
- **Wire format**: msgspec rejects unions of two untagged Structs, so the spec's literal `payload: _Definition | _Empty` field type isn't expressible. Resolved by giving each command a single payload struct with all fields optional — empty wire `{}` decodes to all-None (handler reads as resurrect/no-op intent), populated wire decodes with create-shape fields set. The activate-field handlers additionally validate that all required create fields are present before touching the DB. Spec updated in-place to record this resolution.
- **`ValidationException` handler** added alongside the documented `msgspec.ValidationError` one because Litestar wraps msgspec errors before exception handlers see them.

## Known gaps (deferred)

- `clear_*_field` records the change-log row but does not yet wipe `*_field_values` rows or strip the field key from `properties` JSON — gated on the data-projection spec landing.
- `delete_*_type` cascade removes the schema row + field rows (FK `ondelete=CASCADE`) but does not yet wipe entity rows / their values — same data-projection dependency.
- `actor_id` on `schema_change_log` stays NULL until auth lands.
- `schema_changed` WebSocket broadcast is a separate sync-layer spec.

## Test plan

- [x] 114 tests pass (`uv run pytest tests -v`)
- [x] `uvx ty check src/py tests` reports 0 diagnostics
- [x] Decoder tests cover every command's wire shape (32 tests)
- [x] Service round-trip tests for the 4 entity-kind services + change-log service
- [x] Handler tests cover every cell of the validation matrix per entity kind
- [x] End-to-end controller tests cover happy path, name collision (409), missing entity (404), unknown command (400), unknown payload field (400), and rollback-on-4xx
- [x] OpenAPI test asserts all 18 commands appear under `oneOf` and that 4xx responses reference the error envelope
- [ ] Manual smoke against a running server (`uv run litestar run` or `uv run granian ...`) — left for reviewer if desired

### DIFF
--- a/src/py/novamoc/domain/schema/_dispatch.py
+++ b/src/py/novamoc/domain/schema/_dispatch.py
@@ -1,0 +1,54 @@
+"""Command dispatch.
+
+The handler table is enumerated explicitly below. Each command struct
+maps to the function that executes it. Adding a new verb means: write
+the handler in the appropriate ``_handlers/<kind>.py`` module, then add
+one line here. Explicit beats implicit (Zen of Python item 2) — the
+universe of accepted commands is one ``rg``-able place.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from novamoc.domain.schema import _payloads
+from novamoc.domain.schema._bundle import Handler, ServiceBundle
+from novamoc.domain.schema._handlers import (
+    asset_type,
+    asset_type_field,
+    maintenance_record_type,
+    maintenance_record_type_field,
+)
+from novamoc.domain.schema._outcomes import SchemaCommitOutcome
+
+__all__ = ("dispatch",)
+
+
+_HANDLERS: dict[type, Handler] = {
+    _payloads.CreateAssetType:                      asset_type.create,
+    _payloads.ActivateAssetType:                    asset_type.activate,
+    _payloads.UpdateAssetType:                      asset_type.update,
+    _payloads.DeactivateAssetType:                  asset_type.deactivate,
+    _payloads.DeleteAssetType:                      asset_type.delete,
+    _payloads.CreateAssetTypeField:                 asset_type_field.create,
+    _payloads.ActivateAssetTypeField:               asset_type_field.activate,
+    _payloads.UpdateAssetTypeField:                 asset_type_field.update,
+    _payloads.DeactivateAssetTypeField:             asset_type_field.deactivate,
+    _payloads.ClearAssetTypeField:                  asset_type_field.clear,
+    _payloads.DeleteAssetTypeField:                 asset_type_field.delete,
+    _payloads.CreateMaintenanceRecordType:          maintenance_record_type.create,
+    _payloads.ActivateMaintenanceRecordType:        maintenance_record_type.activate,
+    _payloads.UpdateMaintenanceRecordType:          maintenance_record_type.update,
+    _payloads.DeactivateMaintenanceRecordType:      maintenance_record_type.deactivate,
+    _payloads.DeleteMaintenanceRecordType:          maintenance_record_type.delete,
+    _payloads.CreateMaintenanceRecordTypeField:     maintenance_record_type_field.create,
+    _payloads.ActivateMaintenanceRecordTypeField:   maintenance_record_type_field.activate,
+    _payloads.UpdateMaintenanceRecordTypeField:     maintenance_record_type_field.update,
+    _payloads.DeactivateMaintenanceRecordTypeField: maintenance_record_type_field.deactivate,
+    _payloads.ClearMaintenanceRecordTypeField:      maintenance_record_type_field.clear,
+    _payloads.DeleteMaintenanceRecordTypeField:     maintenance_record_type_field.delete,
+}
+
+
+async def dispatch(services: ServiceBundle, request: Any) -> SchemaCommitOutcome:
+    return await _HANDLERS[type(request)](services, request)
--- a/src/py/novamoc/domain/schema/_handlers/__init__.py
+++ b/src/py/novamoc/domain/schema/_handlers/__init__.py
@@ -1,0 +1,8 @@
+"""Per-entity-kind command handlers.
+
+Each submodule (``asset_type``, ``asset_type_field``, ...) exposes
+module-level handler functions named after the verb (``create``,
+``activate``, ``update``, ``deactivate``, ``clear``, ``delete``). The
+dispatch table in :mod:`novamoc.domain.schema._dispatch` wires those
+functions to their command structs.
+"""
--- a/src/py/novamoc/domain/schema/_handlers/asset_type.py
+++ b/src/py/novamoc/domain/schema/_handlers/asset_type.py
@@ -1,0 +1,140 @@
+"""AssetType command handlers.
+
+Each handler reads the projection, validates the transition, mutates the
+projection (``auto_commit=False``), and appends a ``schema_change_log``
+row. A successful return yields a :class:`SchemaCommitOutcome` whose
+``schema_version`` is the appended row's ``seq``.
+
+Per ADR-008 ``create`` and ``activate`` are separate verbs: ``create``
+takes a full payload and inserts a new row; ``activate`` takes ``{}``
+and only flips ``active = true`` on an existing row.
+"""
+
+from __future__ import annotations
+
+import msgspec
+from advanced_alchemy.exceptions import IntegrityError
+
+from novamoc.domain.schema._commands import SchemaCommand
+from novamoc.domain.schema._errors import (
+    ConflictError,
+    EntityNotFoundError,
+    ErrorCode,
+    PayloadShapeError,
+)
+from novamoc.domain.schema._bundle import ServiceBundle
+from novamoc.domain.schema._outcomes import Outcome, SchemaCommitOutcome
+from novamoc.domain.schema import _payloads
+
+
+async def create(services: ServiceBundle, req: _payloads.CreateAssetType) -> SchemaCommitOutcome:
+    try:
+        await services.asset_type.create(
+            data={
+                "tenant_id": req.tenant_id,
+                "id": req.entity_id,
+                "name": req.payload.name,
+                "active": True,
+            },
+            auto_commit=False,
+        )
+    except IntegrityError as exc:
+        # PK collision on (tenant_id, id) or UNIQUE on (tenant_id, name).
+        raise ConflictError(code=ErrorCode.NAME_RESERVED, name=req.payload.name) from exc
+    row = await services.change_log.append(
+        tenant_id=req.tenant_id,
+        command=SchemaCommand.CREATE_ASSET_TYPE,
+        entity_id=req.entity_id,
+        payload=msgspec.to_builtins(req.payload),
+    )
+    return SchemaCommitOutcome(row.seq, req.entity_id, Outcome.CREATED, row.committed_at)
+
+
+async def activate(services: ServiceBundle, req: _payloads.ActivateAssetType) -> SchemaCommitOutcome:
+    obj = await services.asset_type.get_one_or_none(tenant_id=req.tenant_id, id=req.entity_id)
+    if obj is None:
+        raise EntityNotFoundError(code=ErrorCode.ENTITY_NOT_FOUND)
+    if obj.active:
+        outcome = Outcome.NOOP
+    else:
+        await services.asset_type.update(
+            data={"active": True},
+            item_id=(req.tenant_id, req.entity_id),
+            auto_commit=False,
+        )
+        outcome = Outcome.ACTIVATED
+    row = await services.change_log.append(
+        tenant_id=req.tenant_id,
+        command=SchemaCommand.ACTIVATE_ASSET_TYPE,
+        entity_id=req.entity_id,
+        payload={},
+    )
+    return SchemaCommitOutcome(row.seq, req.entity_id, outcome, row.committed_at)
+
+
+async def update(services: ServiceBundle, req: _payloads.UpdateAssetType) -> SchemaCommitOutcome:
+    obj = await services.asset_type.get_one_or_none(tenant_id=req.tenant_id, id=req.entity_id)
+    if obj is None:
+        raise EntityNotFoundError(code=ErrorCode.ENTITY_NOT_FOUND)
+    # ``omit_defaults=True`` on the payload struct drops UNSET fields here;
+    # an explicit ``null`` on the wire stays as ``None`` so the handler can
+    # write NULL to a nullable column.
+    payload = msgspec.to_builtins(req.payload)
+    if not payload:
+        raise PayloadShapeError(code=ErrorCode.PAYLOAD_NO_CHANGES)
+    try:
+        await services.asset_type.update(
+            data=payload, item_id=(req.tenant_id, req.entity_id), auto_commit=False,
+        )
+    except IntegrityError as exc:
+        # Note: this maps every IntegrityError to NAME_RESERVED. The current
+        # schema only has UNIQUE constraints reachable from update payloads, so
+        # this is correct in practice. If schema columns ever gain CHECK / NOT
+        # NULL constraints reachable from update, distinguish error causes by
+        # inspecting the underlying constraint name.
+        raise ConflictError(code=ErrorCode.NAME_RESERVED) from exc
+    row = await services.change_log.append(
+        tenant_id=req.tenant_id,
+        command=SchemaCommand.UPDATE_ASSET_TYPE,
+        entity_id=req.entity_id,
+        payload=payload,
+    )
+    return SchemaCommitOutcome(row.seq, req.entity_id, Outcome.UPDATED, row.committed_at)
+
+
+async def deactivate(services: ServiceBundle, req: _payloads.DeactivateAssetType) -> SchemaCommitOutcome:
+    obj = await services.asset_type.get_one_or_none(tenant_id=req.tenant_id, id=req.entity_id)
+    if obj is None:
+        raise EntityNotFoundError(code=ErrorCode.ENTITY_NOT_FOUND)
+    if obj.active:
+        await services.asset_type.update(
+            data={"active": False},
+            item_id=(req.tenant_id, req.entity_id),
+            auto_commit=False,
+        )
+        outcome = Outcome.DEACTIVATED
+    else:
+        outcome = Outcome.NOOP
+    row = await services.change_log.append(
+        tenant_id=req.tenant_id,
+        command=SchemaCommand.DEACTIVATE_ASSET_TYPE,
+        entity_id=req.entity_id,
+        payload={},
+    )
+    return SchemaCommitOutcome(row.seq, req.entity_id, outcome, row.committed_at)
+
+
+async def delete(services: ServiceBundle, req: _payloads.DeleteAssetType) -> SchemaCommitOutcome:
+    obj = await services.asset_type.get_one_or_none(tenant_id=req.tenant_id, id=req.entity_id)
+    if obj is None:
+        raise EntityNotFoundError(code=ErrorCode.ENTITY_NOT_FOUND)
+    await services.asset_type.delete(
+        item_id=(req.tenant_id, req.entity_id), auto_commit=False,
+    )
+    row = await services.change_log.append(
+        tenant_id=req.tenant_id,
+        command=SchemaCommand.DELETE_ASSET_TYPE,
+        entity_id=req.entity_id,
+        payload={},
+    )
+    return SchemaCommitOutcome(row.seq, req.entity_id, Outcome.DELETED, row.committed_at)
--- a/src/py/novamoc/domain/schema/_handlers/asset_type_field.py
+++ b/src/py/novamoc/domain/schema/_handlers/asset_type_field.py
@@ -1,0 +1,170 @@
+"""AssetTypeField command handlers.
+
+Per ADR-008 ``create`` and ``activate`` are separate verbs: ``create``
+takes a full payload (asset_type_id, name, data_type, optional
+validation) and inserts a new row; ``activate`` takes ``{}`` and only
+flips ``active = true`` on an existing row. ``create`` validates that
+the parent asset_type exists; a deactivated parent is allowed.
+
+TODO(#7): The ``clear`` handler appends a change-log row but does
+**not** wipe field values from the data-projection store. That wipe is
+gated on the data-projection spec and will be implemented once the EAV
+projection tables land.
+"""
+
+from __future__ import annotations
+
+import msgspec
+from advanced_alchemy.exceptions import IntegrityError
+
+from novamoc.domain.schema._commands import SchemaCommand
+from novamoc.domain.schema._errors import (
+    ConflictError,
+    EntityNotFoundError,
+    ErrorCode,
+    PayloadShapeError,
+)
+from novamoc.domain.schema._bundle import ServiceBundle
+from novamoc.domain.schema._outcomes import Outcome, SchemaCommitOutcome
+from novamoc.domain.schema import _payloads
+
+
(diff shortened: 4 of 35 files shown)


shoriminimoe:
### Code review

Found 1 issue:

1. `update_asset_type_field` and `update_maintenance_record_type_field` cannot clear a nullable field to NULL. The update payload structs use `omit_defaults=True` and `validation: dict | None = None`, so a client sending `{"validation": null}` decodes to a struct whose fields are all at their defaults; `msgspec.to_builtins(req.payload, ...)` then drops those fields, the dict-comprehension None-filter is a no-op, and the handler raises `PAYLOAD_NO_CHANGES`. There is no way for a client to explicitly clear `validation` (or any other nullable column) via this endpoint, and no test exercises that path.

https://github.com/shoriminimoe/novamoc/blob/5f761f467457d56291138442aaee4b55a626432f/src/py/novamoc/domain/schema/_handlers/asset_type_field.py#L116-L121

Same bug in the maintenance-record handler:

https://github.com/shoriminimoe/novamoc/blob/5f761f467457d56291138442aaee4b55a626432f/src/py/novamoc/domain/schema/_handlers/maintenance_record_type_field.py#L126-L131

tests/schema/test_change_log_service.py
```diff
@@ -0,0 +1,51 @@
...
+    row = await svc.append(
+        tenant_id="t1",
+        command=SchemaCommand.ACTIVATE_ASSET_TYPE,
+        entity_id=eid,
+        payload={"name": "Truck"},
+    )
+    await session.flush()
+    assert row.seq is not None
+    assert row.tenant_id == "t1"
+    assert row.command == "activate_asset_type"
```
shoriminimoe: Assert against the enum here instead of the assumed value.

tests/schema/test_entity_services.py
```diff
@@ -0,0 +1,70 @@
...
+    assert fetched is not None
+    assert fetched.name == "X"
+    assert obj.id == eid
+
+
+@pytest.mark.parametrize(
+    ("type_svc_cls", "field_svc_cls", "parent_fk"),
+    [
+        (AssetTypeService, AssetTypeFieldService, "asset_type_id"),
+        (MaintenanceRecordTypeService, MaintenanceRecordTypeFieldService, "maintenance_record_type_id"),
```
shoriminimoe, line 39: Create an issue stating that the *Field classes could be simplified if their parent key were generically names to something like `parent_id``` instead of the specific table. The relationship is part of the ddl and the table name implies the source.

tests/schema/test_dispatch.py
shoriminimoe, line 1: This module should be removed. The tests help with initial development but they will become a nuisance as things change.

src/py/novamoc/domain/schema/_dispatch.py
```diff
@@ -0,0 +1,39 @@
...
+"""Populated by ``novamoc.domain.schema._handlers``. Imports are at the
+bottom of this module so the table is built once at import time."""
+
+
+async def dispatch(services: ServiceBundle, request: Any) -> SchemaCommitOutcome:
+    return await _HANDLERS[type(request)](services, request)
+
+
+# Late import — populating ``_HANDLERS`` by side effect.
+from novamoc.domain.schema import _handlers  # noqa: E402, F401
```
shoriminimoe: Following the Zen of Python in this case, explicit is better than implicit. The handlers should be explicitly registered in the handlers dict. It is not critcal to have the bindings local to the handler definitions.

tests/schema/test_errors.py
```diff
@@ -0,0 +1,49 @@
...
+from novamoc.domain.schema._errors import (
+    ConflictError,
+    EntityNotFoundError,
+    ErrorCode,
+    PayloadShapeError,
+    SchemaCommandError,
+)
+
+
+def test_error_code_membership() -> None:
```
shoriminimoe: Testing the enum values like this is not valuable. The enums themselves are an implementation detail. I would rather rely on tests that exercise user paths and assert on the value and structure there. For example, making a request that causes an error then making an assertion on the output.

tests/schema/test_outcomes.py
shoriminimoe, line 1: This module should be removed.

src/py/novamoc/domain/schema/_payloads.py
shoriminimoe, line 1: Rather than use a custom "command" tag use the msgspec default "type" tag as described in https://jcristharif.com/msgspec/structs.html#tagged-unions. We should also use a base class for the union to make it easier to add new payloads.

To verify this works, the _only_ thing that should change in the payloads tests is the "command" key is changed to "type".

tests/schema/test_controller_unit.py
shoriminimoe, line 1: What is the utility of these tests?

tests/test_smoke.py
shoriminimoe, line 1: This module can be removed.

src/py/novamoc/domain/schema/controllers/_schema.py
```diff
@@ -0,0 +1,128 @@
...
+def schema_command_error_handler(
+    _request: Request[Any, Any, Any] | None, exc: SchemaCommandError,
+) -> Response[dict[str, Any]]:
+    body: dict[str, Any] = {
+        "error": exc.error,
+        "code": exc.code.value,
+        "message": exc.message,
+    }
+    body.update(exc.extras)
+    return Response(content=body, status_code=exc.status_code)
```
shoriminimoe, line 49: Does litestar provide a builtin way for provide error responses?

tests/schema/test_openapi.py
shoriminimoe, line 1: This module can be removed.

tests/schema/test_endpoint_e2e.py
```diff
@@ -0,0 +1,129 @@
...
+            "tenant_id": _T,
+            "entity_id": str(uuid4()),
+            "payload": {},
+        },
+    )
+    assert resp.status_code == 400
+    assert resp.json()["code"] == "invalid_payload_shape"
+
+
+async def test_post_schema_returns_400_on_payload_with_unknown_field(client) -> None:
```
shoriminimoe, line 77: This fails validation by msgspec. Does the error message provide a code that we can assert on? Our client error response body should be consistent and our tests should account for that.

tests/schema/test_handlers_asset_type.py
```diff
@@ -0,0 +1,214 @@
...
+        services,
+        UpdateAssetType(tenant_id=_T, entity_id=eid, payload=_AssetTypeUpdatePayload(name="Lorry")),
+    )
+    await session.flush()
+    assert out.outcome is Outcome.UPDATED
+    row = await services.asset_type.get_one_or_none(tenant_id=_T, id=eid)
+    assert row is not None and row.name == "Lorry"
+
+
+async def test_update_when_tombstoned_is_allowed(session: AsyncSession, services: ServiceBundle) -> None:
```
shoriminimoe: This test should also assert that the updates made it to the database.

src/py/novamoc/domain/schema/_handlers/asset_type_field.py
```diff
@@ -0,0 +1,198 @@
...
+    if obj is None:
+        if empty:
+            raise ConflictError(code=ErrorCode.DEFINITION_REQUIRED)
+        # Validate all required create fields are present before any I/O.
+        if req.payload.asset_type_id is None:
+            raise PayloadShapeError(code=ErrorCode.MISSING_REQUIRED_FIELD, field="asset_type_id")
+        if req.payload.name is None:
+            raise PayloadShapeError(code=ErrorCode.MISSING_REQUIRED_FIELD, field="name")
+        if req.payload.data_type is None:
+            raise PayloadShapeError(code=ErrorCode.MISSING_REQUIRED_FIELD, field="data_type")
```
shoriminimoe: This pattern strongly suggests to me that we should separate "create" from "activate" actions. "create" requires a complete payload. "activate" requires to no payload.

src/py/novamoc/domain/schema/_handlers/asset_type_field.py
```diff
@@ -0,0 +1,198 @@
...
+        tenant_id=req.tenant_id,
+        command=SchemaCommand.DEACTIVATE_ASSET_TYPE_FIELD,
+        entity_id=req.entity_id,
+        payload={},
+    )
+    return SchemaCommitOutcome(row.seq, req.entity_id, outcome, row.committed_at)
+
+
+async def _clear(services: ServiceBundle, req: ClearAssetTypeField) -> SchemaCommitOutcome:
+    # TODO: Wipe field values from the data-projection store once EAV projection
```
shoriminimoe: TODO should have an issue reference. Please create an issue to address this.

tests/schema/test_change_log_service.py
shoriminimoe: Already addressed in f55f164 — the assertions now use `SchemaCommand.ACTIVATE_ASSET_TYPE` etc.

tests/schema/test_entity_services.py
shoriminimoe: Filed as #6.

tests/schema/test_dispatch.py
shoriminimoe: Removed in a8565bf along with the other implementation-detail test modules.

src/py/novamoc/domain/schema/_dispatch.py
shoriminimoe: Refactored to an explicit central registry in 9b1fc29. `_dispatch.py` now enumerates the 22 mappings literally; handler modules expose `activate`, `update`, etc. and don't touch `_HANDLERS`.

tests/schema/test_errors.py
shoriminimoe: Removed in a8565bf — error-path coverage now lives in `test_endpoint_e2e.py` (status code + response `code` field).

tests/schema/test_outcomes.py
shoriminimoe: Removed in a8565bf.

src/py/novamoc/domain/schema/_payloads.py
shoriminimoe: Switched to `tag=_snake_tag` on a `_SchemaCommand` base class in d59eb8d. Each command struct now inherits without per-class `tag=` lines; the wire-format discriminator field is `type` and the values stay snake_case. Test diff: only the `"command":` key in JSON bodies became `"type":`.

tests/schema/test_controller_unit.py
shoriminimoe: Removed in a8565bf.

tests/test_smoke.py
shoriminimoe: Removed in a8565bf.

src/py/novamoc/domain/schema/controllers/_schema.py
shoriminimoe: Documented in dd775dd — Litestar's built-in mechanism for custom error envelopes IS the `exception_handlers` mapping we already use. `HTTPException` defaults to `{status_code, detail, extra}`, which doesn't match our `{error, code, message}` envelope; subclassing it and then overriding the default render produces the same code as the current setup, just with a different inheritance. Module docstring now records this.

tests/schema/test_openapi.py
shoriminimoe: Removed in a8565bf.

tests/schema/test_endpoint_e2e.py
shoriminimoe: Tightened in a8565bf — the unknown-field e2e test now asserts on response `error` and `code` (`invalid_payload_shape`).

tests/schema/test_handlers_asset_type.py
shoriminimoe: Strengthened in a8565bf — both type-level `test_update_when_deactivated_is_allowed` tests now read the row back and assert `name` was written and `active` stayed false. Mirror added to the maintenance-record-type sibling test.

src/py/novamoc/domain/schema/_handlers/asset_type_field.py
shoriminimoe: Resolved by ADR-016 (00f09ed). `create_*` and `activate_*` are now separate verbs: `create_*` requires a full payload (msgspec enforces required fields at decode → `400 invalid_payload_shape`), `activate_*` takes `_Empty` and only flips an existing-but-deactivated row to active. The partial-create-validation block is gone; the field-level handlers are simpler.

src/py/novamoc/domain/schema/_handlers/asset_type_field.py
shoriminimoe: Filed as #7. TODO comments in both field handlers updated to reference it.

shoriminimoe:
All 15 inline review notes have been addressed (see replies on each thread). Summary of the round of changes:

**Phase 1 — fixes & cleanup** (`a8565bf`)
- The agent review's nullable-NULL-clear bug: `update_*UpdatePayload` now uses msgspec's `UNSET` sentinel for nullable-clearable fields. `{"validation": null}` writes NULL; absent leaves the column untouched. Decoder + handler tests cover both paths.
- Six implementation-detail test modules removed (`test_dispatch`, `test_errors`, `test_outcomes`, `test_controller_unit`, `test_smoke`, `test_openapi`); user-path coverage lives in e2e + handler tests.
- `test_update_when_deactivated_is_allowed` now reads back the row; e2e unknown-field test now asserts on response `code`.

**Phase 2 — architecture** (`9b1fc29`, `d59eb8d`, `dd775dd`, `1d766be`, `00f09ed`)
- Explicit central dispatch table in `_dispatch.py`; handler modules no longer self-register.
- msgspec `type` tag on a `_SchemaCommand` base class with a snake-case callable; tests changed only the `"command"` key → `"type"`.
- Litestar built-in error path: documented why `exception_handlers` is the right hook and `HTTPException` subclassing wouldn't reduce code.
- ADR-016 supersedes the activate-as-create-or-resurrect semantics of ADR-008. `create_*` and `activate_*` are now separate verbs (22 commands total, simpler validation matrix, four error codes removed).

**Phase 3 — tracking** (issues opened)
- #6 — Simplify `*Field` models with generic `parent_id` FK.
- #7 — Wire `clear_*_field` value-wipe when data-projection lands.

Final state: `uvx ruff check src/py tests`, `uvx ty check src/py tests`, and `uv run pytest tests` all green. Smoke tested locally with the new `type` discriminator and the create/activate split — every command exercised end-to-end including the rollback-on-4xx contract.

src/py/novamoc/domain/schema/_payloads.py
```diff
@@ -40,64 +63,69 @@ class _Empty(msgspec.Struct, forbid_unknown_fields=True):
...
     tenant_id: str
     entity_id: UUID
     payload: _Empty
 
 
-class DeleteAssetType(msgspec.Struct, tag="delete_asset_type", tag_field="command"):
+class DeleteAssetType(_SchemaCommand):
     tenant_id: str
     entity_id: UUID
     payload: _Empty
```
shoriminimoe: Can this be _Empty or UNSET?

src/py/novamoc/domain/schema/_payloads.py
shoriminimoe, line 1: In all cases where UUID is used in this module, is it important that they are strictly validated as a UUID? Is it not good enough to accept a string?

src/py/novamoc/domain/schema/_handlers/__init__.py
shoriminimoe, line 1: Don't place implementation in `__init__.py` files.

src/py/novamoc/domain/schema/_handlers/asset_type.py
```diff
@@ -16,71 +16,76 @@
...
-    payload_is_empty = req.payload.name is None
+async def create(services: ServiceBundle, req: CreateAssetType) -> SchemaCommitOutcome:
+    try:
+        await services.asset_type.create(
+            data={
+                "tenant_id": req.tenant_id,
+                "id": req.entity_id,
+                "name": req.payload.name,
+                "active": True,
+            },
```
shoriminimoe, line 38: Can a litestar/advanced alchemy DTO be used for these data constructs?

src/py/novamoc/domain/schema/controllers/_schema.py
```diff
@@ -1,12 +1,23 @@
...
+
+**On the built-in path.** Litestar's mechanism for custom error
+envelopes is the ``exception_handlers`` mapping registered on a
+controller (or app). Subclassing ``litestar.exceptions.HTTPException``
+gets you a default render of ``{status_code, detail, extra}`` — useful
+when the default shape fits, less useful when the envelope must be
+``{error, code, message, ...extras}`` like ours. We keep
+``SchemaCommandError`` as a plain ``Exception`` and let the registered
+handler produce the envelope; this *is* the built-in path, just with a
+custom renderer rather than the default HTTPException one.
```
shoriminimoe, line 20: As a follow-on issue, let's make a decision to use [RFC 9457](https://www.rfc-editor.org/rfc/rfc9457.html) problem details.

src/py/novamoc/domain/schema/_dispatch.py
```diff
@@ -1,39 +1,79 @@
...
+    DeactivateMaintenanceRecordType,
+    DeactivateMaintenanceRecordTypeField,
+    DeleteAssetType,
+    DeleteAssetTypeField,
+    DeleteMaintenanceRecordType,
+    DeleteMaintenanceRecordTypeField,
+    UpdateAssetType,
+    UpdateAssetTypeField,
+    UpdateMaintenanceRecordType,
+    UpdateMaintenanceRecordTypeField,
```
shoriminimoe: Rather than import each item (here and elsewhere), import the module and use it to access the classes.

docs/adr/016-split-create-from-activate.md
shoriminimoe, line 1: I changed my mind about adding a new ADR. Let's update ADR 8 with a minimal set of changes then mark it accepted.

src/py/novamoc/domain/schema/_payloads.py
shoriminimoe: Discussed in the plan: the three-state union (`UNSET` / `None` / `dict`) is intentional — `UNSET` means "field absent on wire (don't update)", `None` means "explicit null on wire (clear the column to NULL)", `dict` means "set this value". Reducing the union loses one of those states. Kept as-is. If the intent was to drop the explicit-null path, that would re-open the original bug Phase 1 fixed.

src/py/novamoc/domain/schema/_payloads.py
shoriminimoe: Kept strict UUID validation per your follow-up: invalid input blocked at the API layer, no action.

src/py/novamoc/domain/schema/_handlers/__init__.py
shoriminimoe: Resolved in 478e514. `ServiceBundle` and `Handler` moved to `src/py/novamoc/domain/schema/_bundle.py`; `_handlers/__init__.py` is now a docstring-only marker. `_dispatch.py` no longer re-exports — every consumer (tests, controller, handler submodules) imports `ServiceBundle` directly from `_bundle`.

src/py/novamoc/domain/schema/_handlers/asset_type.py
shoriminimoe: Per your follow-up — db-layer models stay below the service layer. No DTO/model-instance approach; the dict idiom is preserved.

src/py/novamoc/domain/schema/controllers/_schema.py
shoriminimoe: Filed as issue #8 — API-wide, with the existing `error`/`code`/`message`/extras fields mapped onto RFC 9457 slots (`type` URI, `title`, `status`, `detail`, `instance`, plus extension members) rather than carried alongside. Out of scope for #5.

src/py/novamoc/domain/schema/_dispatch.py
shoriminimoe: Resolved in 478e514 + fdb5d95. `_dispatch.py` now does `from novamoc.domain.schema import _payloads` and references `_payloads.ActivateAssetType` etc. Same idiom applied across handlers, controller, and tests in fdb5d95.

docs/adr/016-split-create-from-activate.md
shoriminimoe: Resolved in fbe451b — ADR-016 deleted, ADR-008 edited in place with minimal changes (six verbs, simplified per-command-semantics block, updated payload-conventions table) and flipped to Accepted. Spec/plan/source references updated to point back at ADR-008.

shoriminimoe:
Round-2 review addressed across four commits:

**`fbe451b`** — ADR-016 deleted; ADR-008 updated in place with minimal changes (six verbs, simplified per-command-semantics, updated payload-conventions table) and flipped to Accepted.

**`478e514`** — `ServiceBundle` and `Handler` moved out of `_handlers/__init__.py` into `src/py/novamoc/domain/schema/_bundle.py`. No re-exports — tests, controller, and handler submodules import directly from `_bundle`. `_dispatch.py` switched to `from novamoc.domain.schema import _payloads` (module import; the dict literal references `_payloads.ActivateAssetType` etc.).

**`fdb5d95`** — Module-import idiom swept across the rest of the codebase: handlers, controller, and tests all import `_payloads` (and the controller imports `services as _services`) and qualify references at the use site rather than bulk-importing 5–22 symbols.

**`29a4544`** — Payload is now optional on the 14 no-payload commands (`activate_*`, `deactivate_*`, `clear_*_field`, `delete_*`). Both `{"type": "...", "tenant_id": "...", "entity_id": "..."}` and `{..., "payload": {}}` are accepted on the wire.

**No code change** for the other three threads:
- The `validation: dict | None | UnsetType = UNSET` union stays — each state is meaningful and reducing it loses one path (kept the explicit clear-to-NULL we just added).
- UUID validation stays strict per your follow-up: invalid input blocked at the API layer.
- DTOs / model-instance creates rejected per your follow-up: db-layer models stay below the service layer.

**Issue #8** opened for the API-wide RFC 9457 migration.

Final state: `uvx ruff check src/py tests`, `uvx ty check src/py tests`, and `uv run pytest tests` all green. Smoke-tested locally with the new optional-payload wire shape.

PR #11, opened 2026-05-02
feat(api): RFC 9457 problem-details error envelope (ADR-016, closes #8)

### LINKED ISSUE #8: Adopt RFC 9457 problem-details for API-wide error envelope
**Scope:** Whole API, not just `POST /schema`. The current schema endpoint is the first concrete instance of the envelope; the migration is a global contract change applied uniformly to every present and future endpoint that returns an error.

**Proposal.** Move from the ad-hoc envelope

\`\`\`json
{"error": "conflict", "code": "name_reserved", "message": "...", "name": "Truck"}
\`\`\`

to RFC 9457 problem-details (\`application/problem+json\`). The current custom fields are **adapted** to fit RFC slots rather than carried alongside as ad-hoc extensions:

| Current envelope field | RFC 9457 slot |
|---|---|
| \`error\` (\`invalid_request\` / \`conflict\` / \`not_found\`) | folded into \`type\` URI as the category segment |
| \`code\` (e.g. \`name_reserved\`) | folded into \`type\` URI as the leaf segment, e.g. \`https://novamoc.example/problems/name_reserved\`; remains the stable thing clients branch on |
| \`message\` | \`detail\` |
| HTTP status code | \`status\` |
| Per-error extras (\`name\`, \`field\`, etc.) | RFC §3.2 extension members — same top-level keys, now formally an RFC extension |
| — (new) | \`title\` — short, fixed string per code |
| — (new) | \`instance\` — opaque per-occurrence URI/UUID for log correlation |

Content-type for 4xx/5xx flips from \`application/json\` to \`application/problem+json\`. 2xx responses are unaffected.

**Why now:** no client SDKs exist yet, so the cleanup window is open. Migrating later means coordinating with downstream consumers.

**Implementation outline:**

1. Define a single \`ProblemDetails\` msgspec struct (or equivalent) with \`type\`, \`title\`, \`status\`, \`detail\`, \`instance\`, plus an \`Extra: Mapping[str, Any]\` slot.
2. Extend \`SchemaCommandError\` (and any future error class) to carry a \`title\` and a stable \`type\` URI; the renderer constructs the envelope.
3. Replace the existing \`SchemaErrorResponse\` and the three exception handlers in \`SchemaController\` with a single rendering layer applied across the API (via Litestar's app-level \`exception_handlers\`).
4. Update OpenAPI \`responses=\` mappings on every error path to reference \`ProblemDetails\`.
5. Update the design spec, ADR-013, and any client docs.

**Refs:** [RFC 9457](https://www.rfc-editor.org/rfc/rfc9457.html). Original review thread: PR #5.

Out of scope for #5; tracked here as a follow-on.

shoriminimoe:
## Summary
- Adopts RFC 9457 problem-details (`application/problem+json`) as the API-wide error envelope per ADR-016. Replaces the ad-hoc `{error, code, message, ...extras}` shape on every 4xx/5xx response.
- Wires Litestar's `ProblemDetailsPlugin` at the app level with `enable_for_all_http_exceptions=True` and an `exception_to_problem_detail_map` that converts `SchemaCommandError`, `msgspec.ValidationError`, and `litestar.exceptions.ValidationException` to `ProblemDetailsException`. Future endpoints add a row to the map; no per-controller exception handlers.
- Introduces a `novamoc.api` package as the home for transport-flavored concerns shared across endpoints (today: the `ProblemDetails` msgspec struct for OpenAPI, three converters, `make_instance()` URN helper). Removes the legacy `SchemaErrorResponse` struct and the three controller-scoped handlers in `SchemaController`.

## What changes on the wire
4xx/5xx responses flip from `application/json` to `application/problem+json`. The `code` field becomes the leaf segment of the `type` URI (`https://novamoc.example/problems/<code>`); HTTP `status` carries the broad category; per-error extras (e.g., the conflicting `name`) ride as RFC §3.2 extension members. New `title` (fixed per code) and `instance` (`urn:uuid:<uuid4>` per occurrence) round out the standard slots. 2xx responses are unaffected.

## Documentation
- New: ADR-016 records the decision.
- Updated: ADR-013 cross-references ADR-016, the schema-endpoint design spec adopts the new envelope, CLAUDE.md describes the app-level rendering layer.

## Test Plan
- [x] `uv run pytest` — 102 tests pass (5 new unit tests for `_errors.py`, 6 unit tests for the converters, 1 new e2e test for `instance` + extension members, 4 e2e tests rewritten for the new shape).
- [x] `uv run ruff check src tests` — clean.
- [x] `uv run ruff format --check src tests` — clean (51 files formatted).
- [x] `uvx ty check src tests` — clean.
- [ ] Optional: smoke-test live with `just serve` and `curl -i -X POST http://127.0.0.1:8000/schema/ -H 'Content-Type: application/json' -d '{"type":"do_a_barrel_roll","tenant_id":"t","entity_id":"01958f3b-3b9f-7d3a-89aa-000000000001","payload":{}}'` — expect HTTP/1.1 400, `Content-Type: application/problem+json`, body with `type`, `title`, `status`, `detail`, `instance`.

Closes #8.


src/py/novamoc/domain/schema/_errors.py
```diff
@@ -12,14 +16,17 @@
 from typing import Any
 
 
+PROBLEM_TYPE_BASE = "https://novamoc.example/problems"
```
shoriminimoe: Rather than putting something possibly real, put a non-reachable uri. Also add a TODO with a new issue to address this later.

This also shouldn't be defined here. It should be at the API level.

src/py/novamoc/api/_problem_details.py
```diff
@@ -0,0 +1,97 @@
...
+    extension members are surfaced through Litestar's `extra` mapping
+    on `ProblemDetailsException` and rendered as additional top-level
+    keys (consumers ignore unknown fields).
+    """
+
+    type: str
+    title: str
+    status: int
+    detail: str
+    instance: str
```
shoriminimoe: Is this class used anywhere other than tests? It seems unnecessary.

src/py/novamoc/domain/schema/_errors.py
shoriminimoe, line 1: Because this module defines domain errors, it should be entirely ignorant of API details like ProblemDetails, HTTP status codes and URIs. Those details should be handled in the error adapters.

tests/schema/test_errors.py
```diff
@@ -0,0 +1,36 @@
...
+    EntityNotFoundError,
+    ErrorCode,
+    PROBLEM_TYPE_BASE,
+    PayloadShapeError,
+    SchemaCommandError,
+)
+
+
+def test_problem_type_base_is_stable_https_uri() -> None:
+    assert PROBLEM_TYPE_BASE == "https://novamoc.example/problems"
```
shoriminimoe: This is not a useful test.

src/py/novamoc/domain/schema/_errors.py
shoriminimoe: Addressed in 09ffc04 + 305d3c1. `PROBLEM_TYPE_BASE` moved to `novamoc/api/_problem_details.py` as `_PROBLEM_TYPE_BASE = "urn:novamoc:problems"` (non-reachable URN, opaque per RFC 9457 §3.1) with a `TODO(#13)` comment. Wire format flips from `https://novamoc.example/problems/<code>` to `urn:novamoc:problems:<code>` — leaf segment unchanged. Issue #13 tracks the eventual swap to a published docs URL.

src/py/novamoc/api/_problem_details.py
shoriminimoe: Investigated in c3858b1. The struct IS used outside tests — `SchemaController` references it from `responses=ResponseSpec(ProblemDetails, ...)`, which produces a `#/components/schemas/ProblemDetails` reference in the generated OpenAPI with named fields (type, title, status, detail, instance). Without it, Litestar would emit a generic shape for `ProblemDetailsException`, which is less useful for client codegen. Sharpened the docstring to make the documentation-only purpose unmistakable, and added `media_type="application/problem+json"` to each `ResponseSpec` so the published OpenAPI now correctly reflects the runtime content-type (it was previously documented as `application/json`).

src/py/novamoc/domain/schema/_errors.py
shoriminimoe: Addressed in 09ffc04. `_errors.py` is now domain-only — `ErrorCode`, `_DEFAULT_MESSAGES`, the four exception classes (with one-line docstrings describing the domain situation, no transport vocabulary). Removed: `PROBLEM_TYPE_BASE`, `_TITLES`, the `status_code` class attributes, the `title` and `type_uri` properties. All of those moved to `novamoc/api/_problem_details.py`: `_TITLES` and `_STATUS_CODES` are dicts keyed by `ErrorCode`, and `_type_uri()` builds the URI. The converter `schema_command_error_to_problem_details` looks them up by `exc.code` instead of reading attributes off the exception.

tests/schema/test_errors.py
shoriminimoe: Dropped in 09ffc04. `test_problem_type_base_is_stable_https_uri` is gone, along with `test_every_error_code_has_a_title`, `test_type_uri_is_problem_base_plus_code`, and `test_subclass_status_codes_unchanged` — all four were API-level concerns that don't belong in `test_errors.py` after the refactor. `test_errors.py` now has just two tests: `test_extras_still_carried` and a parametrized `test_subclasses_are_distinguishable` (parametrized in 7e1d6a4 because ty caught the unparametrized form had unreachable branches). Title/URI/status-code coverage moves to `tests/api/test_problem_details.py` where it belongs.

PR #56, opened 2026-05-05
feat: enforce tenant scoping at the storage layer (#51)

### LINKED ISSUE #51: Reject queries against synced tables that are not scoped by tenant
## Requirement

Cross-tenant isolation is a hard system requirement (ADR-014). Every query against a synced table — schema tables, entity projection tables, the event log, field-value projection tables — **must** include `tenant_id` in its filter. A query that omits `tenant_id` must be rejected at construction or execution time, not silently allowed to scan across tenants.

This is an enforcement gap, not a testing gap. Tests are necessary but not sufficient: a missed scope in production code is a data leak between tenants, and "code review attention" (ADR-14's *Consequences*, third bullet) is too thin a guarantee for an invariant that load-bears the multi-tenancy model.

## Concrete example of the gap

While renaming `TenantContext` → `RequestAuth` (PR #49), a regex sweep across handler tests stripped `tenant_id=_T` from `services.<table>.get_one_or_none(id=...)` calls. Every test passed. The unscoped query `SELECT * FROM <table> WHERE id = <uuid>` returned the correct row purely because:

1. The test DB only ever contains tenant `t1`'s data.
2. UUIDs make `id` collisions across tenants statistically impossible.

The discipline ADR-014 relies on (every query visibly scopes by tenant) silently broke and the test suite reported green. The production code in `domain/schema/_handlers/*.py` and `domain/schema/services/*.py` happens to scope correctly — but nothing in the system would have caught the missed scoping if it had landed there instead.

## Acceptance criteria

* Every query method on the projection services (`AssetTypeService`, `AssetTypeFieldService`, `MaintenanceRecordTypeService`, `MaintenanceRecordTypeFieldService`, `SchemaChangeLogService`, and any future synced-table service) **requires** `tenant_id` as a keyword argument and rejects calls that omit it. Rejection is a typed exception, raised at the call site — not a runtime warning, not a silent full-table scan.
* The event log services and any future EAV / field-value projection services land under the same constraint from day one.
* A cross-tenant isolation test exists that:
  * seeds equivalent rows under two distinct tenants (e.g. `tenant-a` and `tenant-b` each have an asset type named `"Truck"` with their own ids),
  * asserts every read scoped to `tenant-a` returns only `tenant-a`'s rows,
  * asserts every read scoped to `tenant-b` returns only `tenant-b`'s rows,
  * asserts that a read constructed without a tenant scope is *rejected* before reaching the database.
* The test exists regardless of whether v1 ships with one tenant or many. The constraint is structural; the test exists to pin it.

## Implementation candidates

Listed for orientation, not as a binding choice — pick whichever fits novaMOC's service-layer conventions:

* **Service-layer guard (preferred for advanced-alchemy alignment).** Wrap each `SQLAlchemyAsyncRepositoryService` subclass so `get_one_or_none`, `list`, `update`, `delete`, etc. validate `tenant_id` is present in their kwargs and raise a typed exception (`UnscopedQueryError` or similar) if not. The wrapper centralises the discipline and is statically inspectable.
* **SQLAlchemy event listener.** A `before_compile` (or `before_execute`) listener inspects the compiled query against synced-table metadata and rejects executions whose WHERE clause has no `tenant_id` predicate. Catches lower-level SQL escapes (raw `session.execute(select(...))`) that a service-layer guard wouldn't see.
* **Both, layered.** The service guard catches misuse early at the call site (better error messages); the event listener is the backstop for any code that bypasses the service layer.

## Related

* ADR-014 — the row-scoping decision. The *Consequences* section names "code review attention to any query that does not visibly scope by tenant" as the mitigation; this issue replaces that mitigation with mechanical enforcement.
* PR #49 — the regression described above; reverted within the PR, but the test suite passed during the regression window.
* ADR-008 schema endpoint, ADR-011 / ADR-012 event-log + projection — the synced-table inventory in scope.

shoriminimoe:
## Summary

Resolves [#51](https://github.com/shoriminimoe/novamoc/issues/51). Cross-tenant isolation becomes a structural property of the storage layer — not a discipline the call sites have to remember.

- **Three SQLAlchemy event listeners** in `src/py/novamoc/db/_listeners.py`:
  - **Layer 1** (`do_orm_execute`): auto-injects a `tenant_id = <ctx>` predicate on every ORM SELECT against a class with a `tenant_id` column. Has a fallback for aggregate transforms (e.g. `count()`) where `state.all_mappers` is empty — adds the WHERE directly on the Core `Select`. Fail-closed: raises `UnscopedQueryError` when no tenant context is set.
  - **Layer 2** (`before_flush`): walks `session.new`, auto-stamps `tenant_id` from the contextvar if unset, raises `CrossTenantWriteError` on explicit-but-mismatched.
  - **Layer 3** (`before_execute`): backstop for Core-level INSERT/UPDATE/DELETE — rejects DML lacking a `tenant_id` predicate (UPDATE/DELETE) or VALUES key (INSERT). SELECTs are excluded; Layer 1 owns them.
- **`TenantContextMiddleware`** (ASGI) reads `request.auth.tenant_id` and binds it to `current_tenant_id` for the request's duration; tests use a `tenant` pytest fixture and `use_tenant` helper. Stacks after the existing `AuthenticationMiddleware`.
- **`TenantScopedMixin`** (`db/models/_mixins.py`) replaces the old `TenantScopedAuditBase`. Projection tables compose `(TenantScopedMixin, UUIDAuditBase)`; log/EAV tables compose `(TenantScopedMixin, DefaultBase)`. `EventLog` is the lone exception (sole `seq` PK with hand-declared non-PK `tenant_id`) because SQLite blocks `INTEGER PRIMARY KEY AUTOINCREMENT` on composite PKs — documented deviation; the listeners' column-presence heuristic still enforces tenancy on it.
- **Handler call sites** drop redundant `tenant_id=auth.tenant_id` from reads (`get_one_or_none`, `list`) and from `create(data={...})` dicts. They keep it inside `update(item_id=(auth.tenant_id, ...))` and `delete(item_id=(auth.tenant_id, ...))` because the composite PK is the WHERE clause. `SchemaChangeLogService.append`/`current_version` no longer take a `tenant_id` argument; `list_for_tenant` collapses to `list` with `OrderBy` moved to the controller (preserves the strong-ETag byte-equality contract).
- **Cross-tenant isolation test** at `tests/schema/test_cross_tenant_isolation.py` proves the issue's acceptance criteria: 11 tests across `list`, `get_one_or_none`, `count`, `update`, plus three programming-error rejection paths (`UnscopedQueryError` on unscoped reads/creates, `CrossTenantWriteError` on cross-tenant writes).

Spec: \`docs/superpowers/specs/2026-05-05-tenant-scoping-enforcement-design.md\`. Plan: \`docs/superpowers/plans/2026-05-05-tenant-scoping-enforcement.md\`.

## Test plan

- [ ] \`just check\` (lint + format + typecheck + tests) green; 166 tests pass.
- [ ] \`tests/schema/test_cross_tenant_isolation.py\` exercises every service method under both \`t-a\` and \`t-b\` and confirms no leak.
- [ ] \`tests/db/test_layer{1,2,3}_*.py\` cover the listeners' positive and negative cases (auto-inject, fail-closed, skip-filter exec option, cross-tenant write rejection, Core-DML backstop).
- [ ] \`tests/accounts/test_tenant_context_middleware.py\` confirms the middleware sets and resets the contextvar around real HTTP requests.
- [ ] Existing schema endpoint suite (\`tests/schema/test_endpoint_e2e.py\`, \`test_read_endpoint_e2e.py\`, \`test_handlers_*.py\`) still passes after handler/service simplification.


shoriminimoe:
### Code review

Found 5 issues:

1. CLAUDE.md "Schema read endpoint" section is stale on the PR branch — still references `GET /schema/{tenant_id}` (path param), `KNOWN_TENANT_IDS`, `TenantNotFoundError`/`tenant_not_found`, and `current_version(tenant_id=...)`, but the PR drops all of those (route is now `GET /schema`, `config.py` retired the registry, `TenantNotFoundError` is removed, and `current_version()` takes no args). CLAUDE.md is project-load-bearing and explicitly cited as guidance to read before non-trivial changes.

https://github.com/shoriminimoe/novamoc/blob/80d475d47cec562437facef4798641ff2cf78f46/CLAUDE.md#L83-L97

2. CLAUDE.md "Testing conventions" still describes the old fixture pattern (`cache=shared&uri=true`, per-test URI derived from `request.node.nodeid`), but the PR replaces that with a `StaticPool` approach in `tests/conftest.py`.

https://github.com/shoriminimoe/novamoc/blob/80d475d47cec562437facef4798641ff2cf78f46/CLAUDE.md#L96-L110

3. `current_version()` docstring claims tenant scoping comes from `loader_criteria`, but the statement is `select(func.coalesce(func.max(SchemaChangeLog.seq), 0))` — a scalar aggregate. `state.all_mappers` is empty for this case, so `with_loader_criteria` does NOT attach. The tenant filter actually comes from the aggregate fallback path in `_listeners.py::_inject_tenant_filter` (the `get_final_froms()` branch). Misdescribing the mechanism will mislead anyone debugging a Layer 1 failure here.

https://github.com/shoriminimoe/novamoc/blob/80d475d47cec562437facef4798641ff2cf78f46/src/py/novamoc/domain/schema/services/_change_log.py#L45-L54

4. `SchemaController` module docstring claims the `GET /schema` handler receives "DI-injected `RequestAuth`", but `read_snapshot` has no `RequestAuth` parameter — it relies on `TenantContextMiddleware` having set the `current_tenant_id` ContextVar, which Layer 1 reads. Only `apply_command` (POST) reads `request.auth` directly, and even that is a middleware-populated attribute access, not DI injection.

https://github.com/shoriminimoe/novamoc/blob/80d475d47cec562437facef4798641ff2cf78f46/src/py/novamoc/domain/schema/controllers/_schema.py#L1-L20

5. The race-safety warning previously added to `SchemaChangeLogService.append` (PR #30, commit `edaaf29`) was silently removed during this PR's signature refactor (commit `6b257fa`). The race itself is unfixed: read-then-insert for `next_seq` under default `BEGIN DEFERRED` still allows two concurrent same-tenant POSTs to pick the same seq and collide on the composite PK as `IntegrityError → 500`. The warning tied that risk to ADR-004's writer-serialization plan; dropping it loses load-bearing context for the next maintainer without justification.

https://github.com/shoriminimoe/novamoc/blob/80d475d47cec562437facef4798641ff2cf78f46/src/py/novamoc/domain/schema/services/_change_log.py#L26-L44

PR #64, opened 2026-05-11
feat(events): per-event field existence + value-type validation

### LINKED ISSUE #23: M1.4 Per-event schema validation (field existence and value type)
## Scope

For each event, validate that the targeted field exists and that the value's shape matches the field's declared `data_type`. Tombstoned fields (`active=false`) are **still accepted** — ADR-012 explicitly decouples the data fold from schema visibility, so this validation checks *existence + type*, not active state.

## Tasks

- For user-field events: look up `(tenant_id, entity_type, field_id)` in `asset_type_fields` / `maintenance_record_type_fields`. Existence is required; `active` is **not** checked.
- For `col:<name>` events: confirm the named column exists on the entity table.
- Validate `value_json` shape against the field's `FieldDataType` (text, number, boolean, date, etc.).
- Add `ErrorCode.UNKNOWN_FIELD` (`404`) and `ErrorCode.VALUE_TYPE_MISMATCH` (`400`) plus problem-details mappings.

## References

- ADR-005 (schema-as-data)
- ADR-008 (server-authoritative schema)
- ADR-012 (`col:` namespace; data fold decoupled from schema visibility)

## Depends on

- #20 (M1.1)

## Acceptance

- An event for a known field with a type-correct value passes validation.
- An event for a known but tombstoned field is **accepted** (per ADR-012).
- An event targeting an unknown field returns `unknown_field`.
- An event whose value type does not match returns `value_type_mismatch`.

shoriminimoe:
Closes #23.

## Summary
- M1.4 of the event-sync endpoint (closes #23, ADR-005/008/012). After the schema-version gate and HLC drift check, each event's `body.values` is now validated against the tenant's current schema.
- **Key classification.** UUID keys → user fields. `col:<name>` → projection columns. Server-managed columns (`col:type_id`, `col:asset_id`, `col:deleted`, `col:row_state_hlc`) reject as `invalid_payload_shape`. Non-UUID non-col keys reject as `invalid_payload_shape`.
- **Field existence.** `AssetTypeFieldService` / `MaintenanceRecordTypeFieldService` lookup by id, with `parent_id == event.type_id` cross-check. A field under a different type reports `unknown_field` rather than silent acceptance. Tombstoned (`active=false`) fields are still accepted per ADR-012.
- **Value shape.** Pure predicate over `FieldDataType`. `null` always passes (cell-clearing sentinel). `bool` is excluded from numeric/integer types so Python's `bool-is-int` quirk doesn't slip through.
- Two new error codes — `unknown_field` (404) with `family` / `type_id` / `field` extras, and `value_type_mismatch` (400) with `field` / `expected` / `received` extras. Both render through the existing converter via subclass dispatch.
- Problem docs at `docs/problems/{unknown_field,value_type_mismatch}.md`.

> ⚠️ Stacked on #63 (M1.3). Base auto-retargets when #63 merges.

## Test plan
- [x] Validator unit tests: `_json_type_name` labels, `_matches_data_type` for every `FieldDataType` including null-as-wildcard and bool-vs-int discrimination.
- [x] Endpoint tests: known user field type-correct → 202; unknown UUID → 404 unknown_field with envelope; cross-type field reported as unknown; deactivated field still accepted; type mismatch → 400 value_type_mismatch; null always accepted; `col:name` text accepted; `col:name` with int → value_type_mismatch; `col:bogus` → unknown_field; `col:deleted` → invalid_payload_shape; non-UUID non-col → invalid_payload_shape; Deactivated body skips per-field validation.
- [x] `just check` clean — 262 tests, ratchet at 23.


shoriminimoe:
### Code review

PR #66, opened 2026-05-11
feat(events): append accepted events to event_log; per-event outcomes

### LINKED ISSUE #24: M1.5 Append accepted events to event_log with idempotent retry
## Scope

Insert validated events into `event_log`. On a `UNIQUE(tenant_id, hlc)` collision, treat the event as already-applied and skip projection writes. Return per-event outcomes so a batch can have mixed results.

## Tasks

- Insert each accepted event into `event_log` with `auto_commit=False` (the controller-level autocommit handler commits at response time).
- On `IntegrityError` from the unique constraint, mark the event `duplicate` rather than `accepted`.
- Compose a per-event outcome list: `[{hlc, outcome: 'accepted' | 'duplicate' | 'rejected:<code>'}]`. The response body is this list.
- Decide and document the all-or-nothing semantics: a single rejected event must not poison the batch — accepted/duplicate ones still apply. (This is what the per-event outcome list assumes.)

## References

- ADR-011 (`event_log` shape; `UNIQUE(tenant_id, hlc)` for idempotency)

## Depends on

- #20 (M1.1)
- #21 (M1.2 — HLC drift)
- #22 (M1.3 — schema-version gate)
- #23 (M1.4 — schema validation)

## Acceptance

- A fresh batch returns all-`accepted` outcomes; `event_log` row count grows by the batch size.
- The same batch posted again returns all-`duplicate`; row count unchanged.
- A batch with one rejected event still appends the others and returns mixed outcomes.

shoriminimoe:
Closes #24.

## Summary
- M1.5 of the event-sync endpoint (closes #24, ADR-011). Validated events are appended to `event_log`; the response now carries per-event outcomes so a batch can have mixed `accepted` / `duplicate` / `rejected:<code>` results without poisoning unaffected events.
- **Per-event vs batch semantics.** Per-event validation failures (HLC drift, unknown field, type mismatch, malformed HLC) now surface as in-band `rejected:<code>` outcomes — the HTTP envelope stays 202. Batch-level failures (`schema_version_stale`, malformed wire) still return `application/problem+json` with 4xx.
- **Idempotent retry.** A retried event whose `(tenant_id, hlc)` collides with an existing row returns `duplicate`. Inserts run inside `session.begin_nested()` so an `IntegrityError` on one event leaves the surrounding transaction usable for the next; the catch targets `advanced_alchemy.exceptions.IntegrityError` (the wrapper around SQLAlchemy's, which `SQLAlchemyAsyncRepositoryService.create` raises).
- **DI bundling.** Four services + the drift tunable are wrapped in an `AppendDeps` dataclass and provided through a single `Provide(_provide_append_deps)` — the handler signature stays at `(self, data, deps)`, avoiding the PLR0913 ratchet bump from a four-service injection.
- **Storage shape.** Event-grain for now: one `event_log` row per wire-event with the body as a msgspec dict in `value_json` and `field_id=NULL`. The EAV-grain decomposition ADR-012 describes lands with the projection-fold work in M1.6/M1.7 where field-grain rows actually become useful.
- Pre-existing event endpoint tests updated for the new outcome envelope.

> ⚠️ Stacked on #64 (M1.4). Base auto-retargets when #64 merges.

## Test plan
- [x] Scaffold + matching-version path returns 202 with single `accepted` outcome.
- [x] Drift / malformed-HLC / unknown-field / value-type-mismatch surface as `rejected:<code>` (202 envelope).
- [x] Mixed batch (good + bad + good) lists each outcome independently — neighbours not poisoned.
- [x] Full replay of accepted batch returns all-`duplicate`.
- [x] Partial replay returns mixed `duplicate` + `accepted`.
- [x] Rejected event does not consume its HLC: a valid event with the same HLC posted afterwards is `accepted`.
- [x] `schema_version_stale` remains a 409 batch-level rejection.
- [x] `just check` clean — 266 tests, ratchet at 23.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -131,21 +131,26 @@ ETag = `"<schema_version>"` (RFC 7232 quoted decimal); `If-None-Match` matching 
 ## Events endpoint (`POST /events`)
 
 Companion to the schema endpoint and the only fully implemented write
-path for **data** events (ADR-002 / ADR-011 / ADR-013). The controller
-enforces three pre-persistence gates and then routes each event to its
-handler:
+path for **data** events (ADR-002 / ADR-011 / ADR-013). Batch-level
+failures (``schema_version_stale``, malformed body) reject the whole
+submission via ``application/problem+json``; per-event work is atomic
+at the event grain (M1.5) and surfaces as ``accepted`` / ``duplicate``
+/ ``rejected:<code>`` outcomes in the response.
+
+Controller responsibilities (all the controller does):
 
 1. **Schema-version gate** (batch-level, ADR-008 / ADR-009) — the
    batch's ``schema_version`` must equal the tenant's current schema
    version, or the whole batch is rejected as ``schema_version_stale``.
 2. **HLC parse + drift bound** (per-event, ADR-006) — each event's
    ``hlc`` is parsed; an HLC more than ``hlc_drift_limit_seconds`` ahead
-   of server wall time raises ``hlc_drift_exceeded``. Past HLCs are
-   always accepted.
-3. **Per-event handler dispatch** (M1.4) — each event is routed via
-   ``_HANDLERS[(event.family, type(event.body))]``; today the handlers
-   do field-existence + value-shape validation, M1.5+ layers persistence
-   and projection writes into the same cells.
+   of server wall time is rejected as
+   ``rejected:hlc_drift_exceeded``. Past HLCs are always accepted.
+3. **Dispatch + outcome aggregation** — each surviving event is routed
+   via ``_HANDLERS[(event.family, type(event.body))]``. The handler
+   validates and appends to ``event_log``; the controller catches any
+   ``DomainError`` it raises, maps it to ``rejected:<code>``, and
+   aggregates the per-event outcomes into the response.
 
 Pipeline mirrors the schema endpoint's shape:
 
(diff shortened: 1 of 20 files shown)


shoriminimoe:
### Code review

Found 3 issues:

1. **Per-event rejected outcomes silently drop the error extras that the problem docs advertise**

   `_rejected()` maps a `DomainError` to `EventOutcome(hlc=..., outcome="rejected:<code>")`, discarding the exception's rich payload. But `HLCDriftExceededError` carries `hlc`, `drift_seconds`, and `limit_seconds`; `UnknownFieldError` carries `family`, `type_id`, and `field`; `ValueTypeMismatchError` carries `field`, `expected`, and `received`. Clients diagnosing a `rejected:hlc_drift_exceeded` outcome have no way to find out *how many seconds* their clock is off, or which field triggered `rejected:unknown_field`. This makes the rejection reason actionable in name only.

   https://github.com/shoriminimoe/novamoc/blob/1c570cb8262faa5a9116e158dc883d04ba590ade/src/py/novamoc/domain/events/controllers/_events.py#L96-L99

2. **`docs/problems/{hlc_drift_exceeded,unknown_field,value_type_mismatch}.md` describe RFC 9457 extension members that no longer appear in any response for these codes**

   These files (served at `/problems/*.html`) say "The response carries the following extension members: `drift_seconds`, `limit_seconds`…" / "`family`, `type_id`, `field`…" / "`field`, `expected`, `received`…". After M1.5 those codes surface exclusively as `rejected:<code>` strings in `EventOutcome.outcome`; the extension members never reach the wire. A developer looking up `rejected:hlc_drift_exceeded` in the problem docs will read about fields that don't exist in the actual response.

   Relevant prior art: the spec (2026-05-11-events-handler-dispatch-design.md) deferred the outcome shape to M1.5+ and said "handlers return `None` and the controller responds 202 with no body" — there was no spec that defined whether extras would carry through into the outcome envelope, so this appears to be an unresolved design gap rather than an intentional omission.

   https://github.com/shoriminimoe/novamoc/blob/1c570cb8262faa5a9116e158dc883d04ba590ade/docs/problems/hlc_drift_exceeded.md#L1-L15
   
3. HLCDriftExceededError (_errors.py:8-28) is dead code in M1.5 — _process_event calls _rejected(event, ErrorCode.HLC_DRIFT_EXCEEDED) directly without raising the typed exception. Meanwhile this PR's own CLAUDE.md update keeps it in the list of per-event errors that "flow through the same problem-details converter as the schema endpoint." Either remove the class and trim the CLAUDE.md list, or raise it in _process_event so the documented contract matches.

PR #67, opened 2026-05-11
feat(events): LWW fold into field-value projection tables

### LINKED ISSUE #25: M1.6 Apply LWW fold to field-value projection tables
## Scope

For each accepted field-grain event, conditionally upsert into the field-value projection table — `asset_field_values` or `maintenance_record_field_values` — only when the new event's HLC strictly beats the stored one. The applied/skipped signal is the gate for #25 (M1.7); without it the entity-table projection can diverge from the field-value projection.

## Tasks

- Apply the upsert:

  ```sql
  INSERT INTO asset_field_values (tenant_id, asset_id, field_id, value_json, hlc)
  VALUES (?, ?, ?, ?, ?)
  ON CONFLICT (tenant_id, asset_id, field_id) DO UPDATE SET
    value_json = excluded.value_json,
    hlc        = excluded.hlc
  WHERE excluded.hlc > asset_field_values.hlc
  RETURNING 1;
  ```

- Capture the `RETURNING` result so the caller can branch ("did this upsert apply?").
- Same shape for `maintenance_record_field_values`.

## References

- ADR-007 (per-field LWW)
- ADR-012 (field-grain projection tables)

## Depends on

- #24 (M1.5)

## Acceptance

- Two events on the same `(entity, field)` posted in HLC order: stored HLC and value match the second event.
- Same posted in reverse order: stored HLC and value still match the higher-HLC event.
- The caller can observe whether an upsert applied or was skipped (RETURNING-based or equivalent).

shoriminimoe:
Closes #25.

## Summary
- M1.6 of the event-sync endpoint (closes #25, ADR-007 / ADR-012). After `event_log` append, each accepted `Created`/`Updated` event's values are folded into `asset_field_values` / `maintenance_record_field_values` via one conditional upsert per `(entity, field)`. The HLC-strict-greater `WHERE` clause is the LWW rule — a stale event observed via catch-up sync cannot overwrite a newer projection value.
- **Applied/skipped signal.** `apply_field_value` returns the `RETURNING` result as a bool. M1.7 will branch on it to gate the entity-table projection update (skipped cell ⇒ stale event ⇒ entity-table column must not be touched, or the two projections diverge).
- **API shape.** `apply_field_value(session, FieldUpsert)` — the bundle keeps the signature at two params and prevents argument transposition. Family routing (asset vs. maintenance_record) lives in a single table-of-truth dict.
- **Atomicity.** The fold runs inside the same `session.begin_nested()` savepoint as the `event_log` insert, so a fold failure rolls back the log row too and the two projections stay consistent.
- **Latent listener bug fixed.** `_values_carries_tenant` was checking `Column.name` on `_values` keys, but SQLAlchemy 2.x normalises `.values(tenant_id=...)` to a string-keyed dict; pre-M1.6 nothing exercised Core-level insert paths, so the bug went unnoticed. Now accepts both shapes.

> ⚠️ Stacked on #66 (M1.5). Base auto-retargets when #66 merges.

## Test plan
- [x] Forward HLC order stores latest value (applied=True both times).
- [x] Reverse HLC order keeps the higher HLC (applied=False on the late arrival).
- [x] Equal HLC does NOT overwrite (strict-greater, projection-level idempotency).
- [x] `null` records as the clear-cell sentinel.
- [x] Maintenance-record family routes to its own table (not asset_field_values).
- [x] Different fields on the same entity are independent.
- [x] `just check` clean — 272 tests, ratchet at 23.

### DIFF
--- a/docs/superpowers/specs/2026-05-12-events-lww-fold-design.md
+++ b/docs/superpowers/specs/2026-05-12-events-lww-fold-design.md
@@ -1,0 +1,323 @@
+# Design: per-field LWW fold for `POST /events`
+
+## Status
+
+Approved 2026-05-12. Implemented in PR #67 (M1.6).
+
+## Problem
+
+M1.5 (PR #66) made the events endpoint durable: each accepted event is
+appended to `event_log` inside a per-event savepoint. But the entity
+projection that clients read (the `*_field_values` EAV table, and
+later the entity-table column / properties JSON cell) is still empty —
+every accepted event survives in the log and not yet anywhere a query
+can see it.
+
+M1.6 closes the first half of that gap. After the `event_log` insert,
+each accepted `Created` / `Updated` event's `values` payload must be
+folded into the per-field projection (`asset_field_values` /
+`maintenance_record_field_values`) under the LWW rule from ADR-007:
+the highest-HLC event wins, ties don't apply, late arrivals lose
+silently.
+
+The fold has three hard constraints that drive the design:
+
+1. **Atomicity with the log.** The two projections (log + EAV) must
+   stay consistent. If the fold fails after the log insert, the log
+   row must roll back too, or a replay would re-fold the same values
+   and corrupt the EAV.
+2. **HLC-strict-greater, in SQL.** Catch-up sync delivers events out
+   of order. A stale event seen after a newer one must not overwrite —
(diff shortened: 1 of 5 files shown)


shoriminimoe:
### Code review

PR #68, opened 2026-05-11
feat(events): mirror applied field events into entity-table projection

### LINKED ISSUE #26: M1.7 Apply entity-table projection updates (properties JSON and col: columns)
## Scope

Mirror each accepted-and-applied field-grain event into the entity table — `properties` JSON for user fields, named columns for `col:` fields. Only write if the M1.6 upsert actually applied; otherwise a stale event would update the entity row without updating the field-value row, diverging the two projections.

## Tasks

- For user-field events: `UPDATE assets SET properties = json_set(properties, '$.<field_id>', <value>) WHERE id = ? AND tenant_id = ?`.
- For `col:<name>` events: `UPDATE assets SET <column> = <value> WHERE id = ? AND tenant_id = ?`.
- Same shapes for `maintenance_records`.
- Clears (`set` with `value_json IS NULL`, or `op='delete'` field-grain events): `json_remove(properties, '$.<field_id>')` for user fields; `SET <column> = NULL` for `col:`.

## References

- ADR-005 (one-row-per-entity + JSON projection)
- ADR-012 (entity-grain projection writes; clears semantics)

## Depends on

- #25 (M1.6)

## Acceptance

- After the M1.6 upsert applies, `json_extract(properties, '$.<field_id>')` (or the named column) returns the same value.
- After a stale (skipped) event, the entity row is unchanged.
- A `set` event with `value_json IS NULL` removes the JSON key (user fields) or nullifies the column (`col:`).
- A field-grain `delete` event has the same effect as a `set` with NULL.

shoriminimoe:
Closes #26.

## Summary
- M1.7 of the event-sync endpoint (closes #26, ADR-005 / ADR-012). Each accepted field-grain event whose M1.6 fold *applied* now mirrors into the entity table — `properties` JSON for user fields, named columns for `col:` fields — so application reads use the entity row directly.
- **Routing.** `col:<name>` → `UPDATE <entity> SET <name> = <value>` (None resolves to SQL NULL, satisfying clears semantics). User-field UUID → `json_set` / `json_remove` on the `properties` column.
- **Gating.** The mirror is gated on the M1.6 `apply_field_value` return value. A stale event whose field-value upsert was rejected by the HLC guard MUST NOT touch the entity row, or the two projections diverge.
- **Partial-pipeline note.** When the entity row hasn't been created yet (the M1.8 row-state path hasn't run), the `UPDATE` silently matches zero rows. M1.7 + M1.8 together close the full apply loop; until then this PR delivers the field-update half. Tests cover that no-op explicitly.

> ⚠️ Stacked on #67 (M1.6). Base auto-retargets when #67 merges.

## Test plan
- [x] `col:name` writes the named column.
- [x] `col:name` value=None nullifies the column.
- [x] User-field writes the key into `properties`.
- [x] User-field value=None removes the key from `properties`.
- [x] User-field update replaces the existing value.
- [x] Two user fields coexist in one `properties` blob.
- [x] Missing-entity-row UPDATE is a silent no-op (M1.8 territory).
- [x] Maintenance-record family routes to its own table (asset row untouched).
- [x] `just check` clean — 280 tests, ratchet at 23.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -249,7 +249,7 @@ Prefer the specialized tooling below over generic Bash/Read/Edit when the task f
 
 - ADR-000 — the ADR practice. To start a new ADR, copy `docs/adr/_template.md`; that template carries the section structure, required-vs-optional rules, status lifecycle, and writing guidance. ADRs 001–015 use the older four-section shape and are grandfathered.
 - ADR-001 overall architecture — three components, two transports, two data classes.
-- ADR-002 / ADR-011 / ADR-012 — event-sourced data + append-only EAV log + JSON-projected entity tables.
+- ADR-002 / ADR-011 / ADR-012 / ADR-019 — event-sourced data + append-only EAV log + JSON-projected entity tables (ADR-019 revises ADR-012's clears clause: a cleared user field stays in `properties` as JSON `null`).
 - ADR-005 / ADR-008 — schema-as-data, server-authoritative; the meta-schema tables and the command verbs.
 - ADR-006 / ADR-007 — HLC ordering and per-field LWW fold (the same fold runs on server and clients).
 - ADR-013 — HTTP and WebSocket transports both carry the same event protocol.
--- a/docs/adr/012-eav-events-with-json-projections.md
+++ b/docs/adr/012-eav-events-with-json-projections.md
@@ -2,7 +2,7 @@
 
 ## Status
 
-Accepted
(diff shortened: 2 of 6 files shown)


shoriminimoe:
### Code review

Branch was rebased onto current main (`af48cb5`); the M1.7 hook lives in `_bundle.py::append_event` rather than `controllers/_events.py` where the original commit placed it, since the M1.5 refactor (#73) had already moved the savepoint + fold into the bundle.

PR #69, opened 2026-05-11
feat(events): row-level events — existence, delete, restore

### LINKED ISSUE #27: M1.8 Row-level events: existence, delete, and restore
## Scope

Handle events with `field_id IS NULL` — these tombstone or restore an entire entity. Field-grain events still apply regardless of visibility (per ADR-012); this issue is solely about the row-level visibility bit and its HLC guard.

## Tasks

- On row-level `set` (`field_id IS NULL`, `op='set'`): upsert the entity row with `deleted=0`, guarded by `excluded.hlc > row_state_hlc`.
- On row-level `delete`: same guard, set `deleted=1`.
- A row-level `set` with HLC lower than the current `row_state_hlc` is a no-op.
- Reads should filter `WHERE deleted=0` — call this out for downstream consumers (M2 read endpoints) but do not implement read filtering here.

## References

- ADR-012 (row-level events; `deleted` + `row_state_hlc` columns)

## Depends on

- #25 (M1.6 — uses the same HLC-guarded upsert pattern)

## Acceptance

- Row-level `delete` followed by row-level `set` with higher HLC: entity row has `deleted=0` and `row_state_hlc` of the second event.
- A field-grain edit between delete and restore lands in `*_field_values` (test-observable) and is reflected in the restored entity's `properties`.
- A row-level `set` with a stale HLC is a no-op.
- No "restore" op exists; restoration is just a row-level `set`.

shoriminimoe:
Closes #27.

## Summary
- M1.8 of the event-sync endpoint (closes #27, ADR-012). Adds the row-state component of the apply: each wire event's body type maps to one row-state mutation under the strict-greater HLC guard.
- **Body → row-state mapping.** `Created` → INSERT-or-upsert the entity row with `deleted=0` (`type_id` from the envelope; `asset_id` from `body.parent.instance_id` for MR). `Activated` → UPDATE-only `deleted=0` with HLC guard. `Deactivated` → UPDATE-only `deleted=1`. `Updated` → no row-state action.
- **Ordering.** `apply_row_state` runs inside `EventServiceBundle.append_event`'s savepoint BEFORE the M1.6 / M1.7 per-field loop, so a Created event's entity row exists by the time M1.7 mirrors values into it. Otherwise json_set / col-update would UPDATE zero rows and the entity-table projection would silently lose initial values.
- **No restore op.** Restoration is just a row-level `set` (Activated) — explicitly per ADR-012.
- **Read filtering on `WHERE deleted=0`** is downstream consumer territory (M2) and explicitly out of scope here.

## Test plan
- [x] Created inserts the entity row with `deleted=False`, `row_state_hlc=event.hlc`, `properties={}`.
- [x] Deactivate flips `deleted` to True with the new HLC.
- [x] Activate restores `deleted=False` after a Deactivate.
- [x] Stale-HLC Deactivate is a no-op (strict-greater).
- [x] Activate / Deactivate on a missing row is a no-op (UPDATE matches 0).
- [x] Updated never touches `row_state_hlc`.
- [x] Replay Create with a lower HLC is a no-op (ON CONFLICT WHERE filters it).
- [x] MR `Created` without `parent` raises a `ValueError` (wire-shape error).
- [x] MR `Created` with `parent` inserts with the correct `asset_id`.
- [x] `just test-py` clean — 326 tests; `ty check` clean; ratchet at 22.

### DIFF
--- a/src/py/novamoc/domain/events/_bundle.py
+++ b/src/py/novamoc/domain/events/_bundle.py
@@ -27,6 +27,7 @@ from novamoc.domain.events._payloads import (
     Updated,
 )
 from novamoc.domain.events._projection import apply_entity_projection
+from novamoc.domain.events._row_state import apply_row_state
 
 if TYPE_CHECKING:
     from uuid import UUID
@@ -137,6 +138,11 @@ class EventServiceBundle:
                     },
                     auto_commit=False,
                 )
+                # Row-state runs BEFORE per-field application so a
+                # Created event's row exists by the time M1.7 mirrors
+                # values into the entity table. Updated has no
+                # row-state component and this call is a no-op for it.
+                await apply_row_state(session, event)
                 for field_id, value in _values_for_fold(event.body).items():
                     upsert = FieldUpsert(
                         family=event.family,
--- a/src/py/novamoc/domain/events/_row_state.py
+++ b/src/py/novamoc/domain/events/_row_state.py
@@ -1,0 +1,144 @@
+"""Row-level event apply (M1.8, ADR-012).
+
+Row-level events carry ``field_id IS NULL`` and toggle the visibility
+of an entire entity by writing ``deleted`` + ``row_state_hlc`` on the
+entity table. Field-grain events fold into ``*_field_values``
+regardless of visibility (ADR-012's "data fold decoupled from schema
+visibility"); only the row-state bit is touched here.
+
+Wire-event mapping:
+
+* ``Created``  → insert-or-upsert the entity row with ``deleted=0`` and
+  the event's HLC as ``row_state_hlc``. Initial field values land in
+  ``properties`` / named columns by way of M1.6 + M1.7 running over
+  the same wire event afterwards.
+* ``Activated`` → UPDATE-only ``deleted=0`` with HLC guard. Missing
+  rows are a no-op — restoration requires a row that has been seen.
+* ``Deactivated`` → UPDATE-only ``deleted=1`` with HLC guard.
+* ``Updated`` → no row-state action.
+
+The HLC guard is strict-greater (same rule as ADR-007's per-field
+fold) so a stale row-state event observed via catch-up sync cannot
+flip the visibility bit when a newer event has already won.
+"""
+
+from __future__ import annotations
+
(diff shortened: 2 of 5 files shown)


shoriminimoe:
### Code review

Found 1 issue:

1. `_apply_create` raises a bare `ValueError` when a `MAINTENANCE_RECORD` `Created` event has `body.parent is None`. The events controller only catches `DomainError` subclasses (mapping them to per-event `rejected:<code>` outcomes), so this escapes as an unhandled exception → HTTP 500 for the whole batch, instead of the intended `rejected:invalid_payload_shape`. The inline comment says "earlier validation should have caught" it, but no MR handler or validator checks `parent` presence, and `parent: Parent | None = None` is optional on the wire (`_payloads.py`). CLAUDE.md says: "generic shape errors reuse `PayloadShapeError(code=ErrorCode.INVALID_PAYLOAD_SHAPE)`" and "catch ``DomainError`` → ``rejected:<code>``". Fix: raise `PayloadShapeError` here, or add a `parent`-presence guard in `_handlers/maintenance_record.py::created` before `services.append_event(event)`.

https://github.com/shoriminimoe/novamoc/blob/28413c0db36f67541f374a3a7c42e026e9243dae/src/py/novamoc/domain/events/_row_state.py#L77-L86

Controller's `DomainError`-only catch:

https://github.com/shoriminimoe/novamoc/blob/28413c0db36f67541f374a3a7c42e026e9243dae/src/py/novamoc/domain/events/controllers/_events.py#L172-L177

shoriminimoe:
Addressed in 907c8ce.

`_row_state._apply_create` now raises `PayloadShapeError(code=INVALID_PAYLOAD_SHAPE)` for the missing-parent path instead of a bare `ValueError`, so the controller's existing `DomainError` catch maps it to `rejected:invalid_payload_shape` as intended.

Coverage:
- `tests/events/test_row_state.py::test_created_maintenance_record_requires_parent` now asserts the typed error + code.
- New `tests/events/test_endpoint_validation.py::test_mr_created_without_parent_is_invalid_payload_shape` exercises the full `/events` request → rejection path and pins the wire-level outcome shape.

PR #98, opened 2026-05-18
feat(events): GET /events catch-up endpoint (M2.4)

### LINKED ISSUE #34: M2.4 GET /events?since for incremental data catch-up
`GET /events?since=<seq>` streams `event_log` rows for the tenant after the given cursor, in `seq` order. Pairs with M2.3 to support reconnect-after-offline. Same wire shape as the M3 WebSocket payload (server emits the same envelope on both transports).

**Refs.** ADR-011, ADR-013 (HTTP half).
**Depends on.** M1 — needs #24 (event_log writes) to produce content.

## Status

`#24` is **closed** — hard dep satisfied.

## Acceptance criteria

Done when:

- A new HTTP route `GET /events?since=<seq>` exists and returns `event_log` rows for the active tenant with `seq > since`, ordered ascending. Lives on `EventsController` or a sibling.
- Tenant resolved by `TenantContextMiddleware` (ADR-017); no tenant in URL or body.
- `since=0` returns the full per-tenant event stream; `since >= MAX(seq)` returns an empty list.
- Cursor is exclusive (`seq > since`), matching ADR-011's prose.
- Response is bounded; carries either a `next_cursor` / `has_more` flag.
- Each emitted event carries:
  - `seq` (replication cursor, opaque to the client)
  - `schema_version` at acceptance time (per ADR-013's per-event tag, drives client-side gating)
  - `hlc`, `family`, `type_id`, `instance_id`
  - `body`: the same discriminated union as POST `/events` (`Created` / `Updated` / `Deactivated` / `Activated`)
- Wire shape (a "RecordedEvent" envelope) is reusable: it lives in `domain/events/_payloads.py` and is the same struct M3 will emit on the WebSocket. Define it here as the single source of truth for the catch-up + live wire shape.
- Tenant scoping verified by cross-tenant isolation test.
- Tests cover empty stream, single-page, multi-page (cursor handoff), and post-deactivated events. No DB mocks.
- `just check` is green.

## Critical design gap (must be resolved in the spec)

`event_log` does not currently store `type_id`. The columns are `seq`, `tenant_id`, `hlc`, `schema_version`, `table_name`, `entity_id` (= instance_id), `field_id`, `op`, `value_json`, `received_at`. The wire envelope requires `type_id`, so the spec must pick one of:

- **(a) Extend `event_log` with a `type_id` column.** Cleanest; pre-release rules allow breaking changes (per CLAUDE.md), and the column is small. Requires a small write-path update in `EventServiceBundle.append_event`. Recommended.
- **(b) Recover via JOIN to `assets` / `maintenance_records` on `(tenant_id, entity_id)`.** Works while the entity exists in the projection — we have no hard-delete today (only `Deactivated` tombstones), so this is currently safe. Adds JOIN cost per event.
- **(c) Stuff `type_id` into `value_json` on write.** Couples the storage layer to the wire shape; not recommended.

Pick (a) unless the spec turns up a concrete reason not to.

## Open design questions

- Batch size cap and how it's configured (Settings vs. hardcoded).
- Whether to short-circuit when the client's `schema_version` (if supplied) is stale — or always emit events tagged with their acceptance-time `schema_version` and let the client gate (ADR-013's recommended pattern). Latter is simpler.

## Implementation workflow

1. **Brainstorm** (`superpowers:brainstorming`) — `type_id` recovery is the load-bearing question.
2. **Write a spec** under `docs/superpowers/specs/<date>-events-catchup-design.md`. The spec must take a position on `type_id` storage and on the shared catch-up/WS envelope shape.
3. **Write an execution plan** under `docs/superpowers/plans/<date>-events-catchup.md`.
4. **TDD** against in-memory fixtures; reuse the events test conventions.
5. **Verify** via `just check`.

shoriminimoe:
## Summary

Closes #34 (M2.4 — GET /events catch-up endpoint, the HTTP half of ADR-013).

Adds the read-side counterpart to `POST /events`: a cursor-paginated HTTP endpoint that returns the active tenant's `event_log` rows after a client-supplied cursor, in `seq` order, in bounded batches. Defines the `RecordedEvent` wire envelope that M3's WebSocket fan-out will reuse verbatim per ADR-013 ("wire format identical regardless of transport").

13 commits: 2 docs (spec + plan) + 10 implementation + 1 final-review cleanup.

## Key changes

- **Storage:** `event_log` gains a `type_id: Mapped[str]` column so envelopes can be reconstructed without joining the projection. Pre-release schema change, no migration tooling needed.
- **Wire envelope:** new `RecordedEvent` msgspec struct in `domain/events/_payloads.py`. Adds `seq` / `schema_version` / `received_at` to the write-side `EventEnvelope` body shape. Same `EventBody` discriminated union, same `EntityFamily`, so clients pattern-match identically regardless of direction.
- **Pagination:** Litestar's `CursorPagination[int, RecordedEvent]` via a new `EventLogCursorPaginator(AbstractAsyncCursorPaginator[...])`. Cursor is raw `seq` (exclusive `seq > cursor`, ADR-011). Fetches `results_per_page + 1` to detect overflow without `COUNT`.
- **Controller:** thin `read_stream` handler on the existing `EventsController` at `GET /events/`. `Parameter(ge=0)` / `Parameter(ge=1, le=5000)` constraints render bad input as `application/problem+json` per ADR-016.
- **Settings:** `EVENT_CATCHUP_DEFAULT_BATCH_SIZE = 500` and `EVENT_CATCHUP_MAX_BATCH_SIZE = 5000` module-level constants in `config.py`, imported directly by the controller (no `AppSettings` indirection — a deliberate change from the original plan; see commit `563d481` for rationale).
- **Schema-version policy:** no short-circuit on read. Events always carry their acceptance-time `schema_version`; clients gate locally per ADR-013.

## Tests

25 new tests, all against the real in-memory SQLite (no DB mocks):

- `type_id` persistence (1)
- `RecordedEvent` round-trip + `body_from_row` per body variant + family inverse-map (7)
- `EventLogCursorPaginator` unit tests (5: empty, under-page-size, signals-more, cursor handoff, exact boundary)
- Endpoint E2E (8: empty, happy path, pagination handoff, all 4 body variants, schema_version preservation across schema change, 3 × 400 ProblemDetails)
- Cross-tenant isolation (1: interleaved t-a / t-b, structural Layer-1 scoping)

## Pipeline

- `uv run pytest`: 362 passed
- `uv run ty check`: clean
- `uv run ruff check`: 22 violations matching ratchet baseline (no new debt)
- `just ratchet`: ok

## Test plan

- [ ] Review the spec + plan + cumulative diff
- [ ] Confirm the deliberate plan deviations are sound:
  - `_int_env` and the `AppSettings.event_catchup_*` fields were removed after the controller bound to module constants directly (avoids env-var-coupled import-time assertion brittleness)
  - The cursor-pagination filter uses `EventLog.seq > cursor` as a positional `ColumnElement` predicate rather than a custom `_seq_gt` helper
- [ ] Approve to merge

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -202,6 +202,56 @@ endpoint. Per-event error types live in ``domain/events/_errors.py``
 ``UnknownFieldError``, ``ValueTypeMismatchError``); generic shape errors
 reuse ``PayloadShapeError(code=ErrorCode.INVALID_PAYLOAD_SHAPE)``.
 
+## Events catch-up endpoint (`GET /events`)
+
+Counterpart to ``POST /events`` and the HTTP half of the catch-up
+flow (ADR-013). Returns the active tenant's ``event_log`` rows after
+a client-supplied cursor in ``seq`` order, in bounded batches.
+
+Response shape is Litestar's
+``CursorPagination[int, RecordedEvent]``: ``items`` plus a
+``cursor`` field that echoes back into the next request's
+``?cursor=`` query parameter, or ``null`` when the caller has
+reached the end. Cursor semantics are exclusive (``seq > cursor``,
+ADR-011). ``RecordedEvent`` (``domain/events/_payloads.py``) is the
+read-side twin of ``EventEnvelope`` — it adds the server-assigned
+fields (``seq``, ``schema_version``, ``received_at``) the
+write-side envelope lacks; the ``body`` field is the same
+discriminated union as on the POST. The M3 WebSocket fan-out will
+emit the same struct so the wire format is transport-independent
+(ADR-013).
+
+``event_log.type_id`` is populated on every accepted event so the
+read side can reconstruct the envelope without joining the
+projection. ``body_from_row`` in ``_bundle.py`` is the inverse of
+``_value_json_for_body`` / ``_op_for_body`` — ``DELETE``-op rows
+reconstruct to ``Deactivated()``; every other row decodes via
+``msgspec.convert(value_json, type=EventBody)`` using the ``event``
+discriminator tag.
+
+Batch size: ``cursor`` defaults to ``None`` (start of stream),
+``results_per_page`` defaults to ``EVENT_CATCHUP_DEFAULT_BATCH_SIZE``
+(500) and is capped at ``EVENT_CATCHUP_MAX_BATCH_SIZE`` (5000); both
+constants live in ``config.py`` and are imported directly by the
+controller. Bad input (negative cursor, out-of-range batch size)
+renders as ``application/problem+json`` per ADR-016, via Litestar's
+standard validation pipeline — no new error codes.
+
+The endpoint always returns events with their acceptance-time
+``schema_version`` tag and does **not** short-circuit on client
+staleness (ADR-013 §"Schema version tagging on events"). Clients
+gate locally — any event with
+``schema_version > active_schema_version`` is buffered until the
+client transitions through the upgrade flow (ADR-009).
+
+Implementation lives in ``domain/events/_pagination.py``
+(``EventLogCursorPaginator`` extends Litestar's
+``AbstractAsyncCursorPaginator[int, RecordedEvent]``); the
+controller's ``read_stream`` handler is a thin pass-through. Tenant
+scoping is structural — Layer 1 of ``db._listeners`` injects the
+``WHERE tenant_id = <ctx>`` predicate on every ORM SELECT, so the
+paginator carries no tenant predicate of its own.
+
 ## Data model conventions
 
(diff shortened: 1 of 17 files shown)


shoriminimoe:
### Code review

Found 1 issue:

1. The new `@get`-decorated `read_stream` handler suppresses TC001/TC002 with per-line `# noqa` on `CursorPagination` and `RecordedEvent` because they're used in the handler's return annotation. CLAUDE.md says: *"Prefer a targeted config knob over an ignore when one applies — the rule keeps firing where it should, just not where the configured framework legitimately violates it."* `pyproject.toml` already lists `"litestar.post"` in `runtime-evaluated-decorators` for the exact same reason (handler decorators introspect signatures at request time); add `"litestar.get"` alongside it and drop the two new `noqa` comments.

Controller noqa lines:
https://github.com/shoriminimoe/novamoc/blob/352844bc1c56737ab9aa3272cf62760bedc50f8a/src/py/novamoc/domain/events/controllers/_events.py#L43-L70

Existing `runtime-evaluated-decorators` config that should be extended:
https://github.com/shoriminimoe/novamoc/blob/352844bc1c56737ab9aa3272cf62760bedc50f8a/pyproject.toml#L137-L146

PR #101, opened 2026-05-18
feat(schema): GET /schema/changes catch-up endpoint (#32)

### LINKED ISSUE #32: M2.2 GET /schema/changes?since for incremental schema catch-up
`GET /schema/changes?since=<seq>` streams ordered `schema_change_log` rows for the tenant after the given seq. Lets a returning client step forward through each schema mutation rather than re-fetching the whole snapshot.

**Refs.** ADR-008, ADR-009.
**Depends on.** None (independent of M2.1 — different shape).

## Status

Soft-dep on #17 (per-tenant dense `seq`) is **satisfied** — #17 closed.

## Acceptance criteria

Done when:

- A new HTTP route `GET /schema/changes?since=<seq>` exists on `SchemaController` and returns the active tenant's `schema_change_log` rows with `seq > since` ordered ascending.
- Tenant is resolved by `TenantContextMiddleware` (ADR-017); no tenant ever appears in the URL or body. Layer 1 of `db._listeners` supplies the WHERE predicate.
- `since=0` returns the full per-tenant schema history; `since >= current_version` returns an empty list (not an error).
- Cursor semantics are exclusive (`seq > since`), matching ADR-011's prose.
- Response is bounded — a single request returns at most a configurable batch size, and the wire shape signals whether more rows remain (e.g. `next_cursor` / `has_more`). Pick the shape during design.
- Each row carries at minimum: `seq`, `command`, `entity_id`, `payload` (the original POST body), `committed_at`. `actor_id` is included once auth is real (M5) — for now it can be omitted or always-`null`.
- Wire shape lives in `domain/schema/_read_payloads.py` (or sibling) as msgspec Structs; types are publishable in OpenAPI.
- Tenant scoping verified by a cross-tenant isolation test (parallel to `tests/schema/test_cross_tenant_isolation.py`).
- Tests follow the schema endpoint convention: E2E HTTP test for status/wire shape + handler/service-level test for query behaviour.
- `just check` is green.

## Open design questions (resolve in the spec)

- Cursor + batch-size contract: opaque `next_cursor` token, or just echo `next_since` integer? Bounded batch size: hardcode or surface via `Settings`?
- Should `payload` round-trip back through the original `SchemaCommand` struct for validation on emission, or be emitted as-is from `JsonB`? The latter is simpler; the former catches drift.

## Implementation workflow

1. **Brainstorm** (`superpowers:brainstorming`) the open design questions above.
2. **Write a spec** under `docs/superpowers/specs/<date>-schema-changes-catchup-design.md`.
3. **Write an execution plan** under `docs/superpowers/plans/<date>-schema-changes-catchup.md` (`superpowers:writing-plans`).
4. **Implement test-first** (`superpowers:test-driven-development`) against the in-memory DB fixtures — no DB mocks.
5. **Verify before completion** (`superpowers:verification-before-completion`): `just check` must pass.

shoriminimoe:
## Summary

Implements `GET /schema/changes?since=<seq>&limit=<n>` per the design at `docs/superpowers/specs/2026-05-18-schema-changes-catchup-design.md` — the server side of ADR-009's catch-up flow. A returning client steps forward through accepted schema commands rather than re-fetching the full snapshot.

- **Echo cursor** (`next_since` + `has_more`) rather than an opaque token — `seq` is already publicly meaningful as `schema_version`, and ADR-008/ADR-009 explicitly describe the protocol in terms of `seq > V_old`.
- **Payload passed through as-is** from `JsonB`. Round-tripping through command-side `_payloads.py` structs would couple history reads to current struct shapes (renames break old rows) without adding safety — the payload was already validated at POST time.
- **Batch size cap** as `AppSettings.schema_changes_max_batch_size` (default 500, env override `NOVAMOC_SCHEMA_CHANGES_MAX_BATCH_SIZE`). Same DI pattern `EventsController` uses for `hlc_drift_limit_seconds`.
- **Tenant scoping** is structural — handler takes no `tenant_id`; Layer 1 of `db._listeners` supplies the `WHERE` predicate. Cross-tenant isolation test added.
- **Range errors** (`since < 0`, `limit` out of bounds, non-integer query) render via the existing `INVALID_PAYLOAD_SHAPE` problem-details path. No new `ErrorCode` values.
- **Snapshot consistency** — `current_version()` and the page query share the request-scoped session, so they observe one SQLite WAL snapshot.

Refs ADR-008 (server-authoritative schema), ADR-009 (mandatory schema upgrade), ADR-016 (RFC 9457 problem-details), ADR-017 (tenant resolution).

Closes #32.

## Test plan

- [x] `tests/schema/test_changes_endpoint_e2e.py` — 13 E2E tests: empty tenant, cursor semantics (`since=0` / `since>=current` / exclusive lower bound), paging, row shape (`actor_id: null`, payload pass-through, `committed_at` ISO-8601), deactivate→activate as separate rows, 400 range errors, 401 missing-auth.
- [x] `tests/schema/test_changes_service.py` — 5 service-level tests covering `list_changes_after` ordering, limit, exclusive-since, empty-tenant behaviour.
- [x] `tests/schema/test_cross_tenant_isolation.py` — adds parametrized test verifying `list_changes_after` returns only the active tenant's rows even when sibling tenants have interleaved `seq` values.
- [x] `tests/test_config.py` — extends the existing config suite with `_int_env` coverage and the new `schema_changes_max_batch_size` field (default / env override / junk-rejection).
- [x] `just check` — 366 Python tests pass, lint/format/typecheck clean, ratchet matches baseline. (`test-js-unit` fails with `vitest: not found` — pre-existing environment gap on this worktree, unrelated to these changes.)

### DIFF
--- a/docs/superpowers/plans/2026-05-18-schema-changes-catchup.md
+++ b/docs/superpowers/plans/2026-05-18-schema-changes-catchup.md
@@ -1,0 +1,985 @@
+# Schema Changes Catch-up Endpoint Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Implement `GET /schema/changes?since=<seq>&limit=<n>` per the design at `docs/superpowers/specs/2026-05-18-schema-changes-catchup-design.md` — a bounded, transactionally-consistent stream of accepted schema commands for the active tenant, the server side of ADR-009's catch-up flow.
+
+**Architecture:** New `@get("/changes")` handler on the existing `SchemaController`. Reuses `SchemaChangeLogService.current_version()` for the snapshot read; adds one `list_changes_after(since, limit)` method to the same service for the page. Response shape extends `_read_payloads.py` with a new `SchemaChangeView` row struct and a `SchemaChangesResponse` envelope. Batch-size cap lives as an `AppSettings` field, injected via the same `Provide(_provide_<name>)` pattern `EventsController` already uses. Bounds errors raise `PayloadShapeError(code=INVALID_PAYLOAD_SHAPE)` — no new `ErrorCode` values.
+
+**Tech Stack:** Python 3.14, Litestar, msgspec, advanced-alchemy + SQLAlchemy 2 (async), aiosqlite, pytest (asyncio auto mode), uv, ruff, ty.
+
+---
+
+## File map
+
+**Created:**
+- `tests/schema/test_changes_endpoint_e2e.py` — E2E HTTP tests (200 empty / 200 populated / paging / cursor semantics / range errors / auth).
+- `tests/schema/test_changes_service.py` — service-level tests for `list_changes_after`.
+- `tests/test_config.py` (new) or extend an existing test — env-var parsing for the new `_int_env` and `schema_changes_max_batch_size` field. **Decision:** colocate in a new `tests/test_config.py`. (No existing file covers `config.py`; this is the first one.)
+
+**Modified:**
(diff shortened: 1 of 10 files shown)


shoriminimoe:
### Code review

PR #104, opened 2026-05-19
M2.3: GET /snapshot bulk projection transfer

### LINKED ISSUE #33: M2.3 GET /sync/initial bulk projection transfer
`GET /sync/initial` streams the current `assets`, `maintenance_records`, and `*_field_values` for the tenant, plus a `cursor` (the current `event_log.seq` for the tenant). The client persists projections + cursor; subsequent reconnects use M2.4 to catch up incrementally.

**Refs.** ADR-015.
**Depends on.** M1 — cursor is read from `event_log`, which only exists once #24 is in.

## Status

`#24` (event_log writes) is **closed** — hard dep satisfied. ADR-015 is still **Proposed**; M2.5 flips it to Accepted once this ships.

## Acceptance criteria

Done when:

- A new HTTP route `GET /sync/initial` exists and returns a paginated bulk transfer of the active tenant's data-projection state, per ADR-015.
- Tenant is resolved by `TenantContextMiddleware` (ADR-017); no tenant in URL or body.
- Each batch carries:
  - Rows from one or more projection tables — `assets`, `maintenance_records`, `asset_field_values`, `maintenance_record_field_values`.
  - The server's current `schema_version` (ADR-015's cross-batch consistency invariant).
  - A continuation token when more batches remain; absence = transfer complete.
  - On the final batch only: the per-tenant cursor (`MAX(event_log.seq) WHERE tenant_id = <ctx>`), used by the client as the starting point for incremental sync (M2.4) and WS hello (M3).
- Each batch is served under one SQLite read snapshot (single transaction), so a batch is internally consistent. Cross-batch consistency is enforced by `schema_version`: if it advances between batches the client restarts (this endpoint enforces by including `schema_version` on every batch — the client makes the decision).
- Tombstones (`active=false` types/fields, `deleted=true` rows) are included so the client can render row-state correctly.
- HLCs on `*_field_values` rows are preserved so the client's subsequent local fold against incoming events works.
- `properties` JSON on entity rows: default to client-computed (omit / compute on client from field values) per ADR-015's recommendation; revisit only if profiling demands it.
- Bounded batch size, configurable via `Settings` (start at a few thousand rows per batch per ADR-015).
- Wire shape lives in a new `domain/sync/` package or as a sibling of the schema/events payloads. Types publish in OpenAPI.
- Tenant scoping verified by cross-tenant isolation test.
- Tests cover: empty tenant (no projections), single-batch transfer, multi-batch transfer, mid-transfer schema-version change (server keeps emitting the new version; client behaviour is its own concern).
- `just check` is green.

## Open design questions (resolve in the spec)

- **Continuation token format**: opaque base64-encoded `(table, last_id)` tuple? Plain `?cursor=` integer per-table? Anything that survives resume across processes.
- **Batch composition**: one table per batch (simpler, predictable), or interleaved up to a row cap (more efficient for small tenants)?
- **Row ordering** within a batch must be stable (`ORDER BY id` per table) so resumption is well-defined.
- **Where to host the route** — new `SyncController`, or fold into one of the existing controllers? The natural place is a new module, since sync is its own concern.

## Implementation workflow

1. **Brainstorm** (`superpowers:brainstorming`) — esp. continuation token + batch composition.
2. **Write a spec** under `docs/superpowers/specs/<date>-sync-initial-design.md`.
3. **Write an execution plan** under `docs/superpowers/plans/<date>-sync-initial.md`.
4. **TDD** against in-memory fixtures, no DB mocks.
5. **Verify** via `just check`.

shoriminimoe:
## Summary
- New `GET /sync/initial` endpoint streams the active tenant's data projection (`assets`, `asset_field_values`, `maintenance_records`, `maintenance_record_field_values`) in fixed-table-order batches with an opaque continuation cursor, plus the per-tenant `event_log.seq` cursor for incremental sync. Closes #33; flips ADR-015 to Accepted.
- Cursor encodes `(start_seq, table, last_id)` as URL-safe base64 JSON. `start_seq` is captured on the **first** request and threaded through every subsequent cursor — this is the correctness pin so events arriving mid-transfer aren't silently dropped. Returned as `event_log_cursor` on the terminal batch.
- `InitialSyncBatch` is a custom envelope (not Litestar's `CursorPagination` — items are heterogeneous across batches): `schema_version`, `cursor`, `event_log_cursor`, and a `body` discriminated union tagged on `table`. Row views omit `name` / `properties` — client recomputes from per-field rows (ADR-015 default).
- New package `domain/sync/` with `_cursor.py`, `_payloads.py`, `services.py` (4 read-only services), `_pagination.py` (the `InitialSyncPaginator` that walks the four tables and collapses empty intermediates server-side), and `controllers/_sync.py`. Tenant scoping is structural via the existing Layer 1 listener.
- Settings: `INITIAL_SYNC_DEFAULT_BATCH_SIZE = 1000`, `INITIAL_SYNC_MAX_BATCH_SIZE = 5000`. Out-of-range page size renders as `application/problem+json` via Litestar's `ValidationException`; malformed cursor as `INVALID_PAYLOAD_SHAPE` via `PayloadShapeError`. No new `ErrorCode`.
- Also adds `EventLogService.current_seq()` (mirror of `SchemaChangeLogService.current_version()`) for the start-of-transfer seq snapshot.

## Test plan
- [x] `tests/sync/test_cursor.py` — 11 cases: roundtrip across every (table, last_id) shape, tamper rejection (bad base64, non-JSON, missing fields, unknown table), padded-token acceptance.
- [x] `tests/sync/test_pagination.py` — 8 paginator scenarios against a real in-memory aiosqlite: empty tenant, single-table-fits-one-page, multi-page cursor handoff, cross-table walk over all four populated, skip-empty-intermediate, **start_seq captured at first request** (correctness pin — insert an event mid-walk, assert the terminal cursor is pre-extra), `schema_version` is current at each request, bad cursor surfaces `PayloadShapeError(INVALID_PAYLOAD_SHAPE)`.
- [x] `tests/sync/test_endpoint_sync_initial.py` — 8 E2E cases through the `client` fixture: empty tenant terminal batch, POST-event + GET round trip with HLC preservation, multi-batch round trip (5 assets, page=2), mid-transfer schema-version observation, bad cursor → 400 problem+json, page-size below min / above max → 400, tombstoned asset surfaces with `deleted=true`.
- [x] `tests/sync/test_sync_cross_tenant_isolation.py` — `t-a` and `t-b` seed parallel data; under each tenant context the paginator returns only that tenant's assets. Belt-and-braces check the other tenant's id is absent.
- [x] `just check`: lint, format, typecheck, ratchet (22 violations, baseline unchanged), **422 python tests pass**. (`test-js-unit` failure is a pre-existing environment issue — `src/js/web/node_modules` is not installed on this branch.)

## Notes for review
- Spec: `docs/superpowers/specs/2026-05-18-sync-initial-design.md`.
- Plan: `docs/superpowers/plans/2026-05-18-sync-initial.md`.
- ADR-015 flips from Proposed → Accepted (2026-05-19).
- One spec amendment landed during impl: empty-tenant terminal batch tags as the *last* table (the algorithm correctly skips empty intermediates server-side), not `assets` as earlier flavor text claimed.
- `InitialSyncPaginator._read_page` uses an OR-expansion for the row-value `>` predicate on `*_field_values` tables (`asset_id > eid OR (asset_id = eid AND field_id > fid)`) rather than `sqlalchemy.tuple_()` — the spec documents this as the portable fallback when the tuple form misbehaves at the typed layer; clean ty under this form.
- `CLAUDE.md` gains an "Initial sync endpoint (`GET /sync/initial`)" subsection documenting the envelope, the cursor format, the start_seq correctness pin, and the empty-intermediate collapse.


shoriminimoe:
### Code review

Found 1 issue:

1. `_SyncBody` is missing `forbid_unknown_fields=True`, so the four `*BatchBody` subclasses silently drop unknown fields. The module docstring claims `forbid_unknown_fields=True` is set "on every struct", and the parallel `_Body` base in `domain/events/_payloads.py` does set it — msgspec does not inherit the option from a parent struct, so each class has to declare it explicitly.

https://github.com/shoriminimoe/novamoc/blob/73360c080f6d7806a7af50441e25244e7b5adb50/src/py/novamoc/domain/sync/_payloads.py#L73-L99

shoriminimoe:
Fixed in e6457d2. Set `forbid_unknown_fields=True` on `_SyncBody` and added `tests/sync/test_payloads.py` parametrised across all four `*BatchBody` variants as a regression guard.

One technical note on the reasoning: msgspec *does* inherit struct config from a parent — I verified with a small repro, and `dir()`-ing a child shows the parent's options carry through. The behaviourally-equivalent diagnosis is "the base wasn't setting it, so the subclasses had nothing to inherit." Fix is the same either way; called it out so the assumption "msgspec doesn't inherit" doesn't carry into future reviews.

shoriminimoe:
Renamed `/sync/initial` → `/snapshot`. Two reasons to revisit the naming after design review:

1. The endpoint is conceptually a *snapshot* of current state (historical/as-of-seq snapshots are out of scope per ADR-015) — `snapshot` is more accurate than `sync`.
2. The previous response shape mixed two cursor concepts under one name. Now disambiguated:
   - `page: str | None` — opaque pagination continuation (was `cursor`). Passed back as `?page=`.
   - `cursor: int | None` — replication `event_log.seq` (was `event_log_cursor`). Present only on the terminal batch.

Mechanical sweep covers the directory rename (`domain/sync/` → `domain/snapshot/`, `tests/sync/` → `tests/snapshot/`), file renames (`_cursor.py` → `_page.py`, `controllers/_sync.py` → `controllers/_snapshot.py`), class/identifier renames (`SyncController`/`InitialSync*` → `SnapshotController`/`Snapshot*`, `CursorState` → `PageState`, `INITIAL_SYNC_*_BATCH_SIZE` → `SNAPSHOT_*_BATCH_SIZE`), handler rename (`initial` → `read`), test renames + assertions updated, spec/plan rename + identifier updates with a rename note at the top, and a rewritten CLAUDE.md subsection.

`just check`: clean (lint, format, typecheck, 22-violation ratchet at baseline, 426 python tests passing). Remote branch name `m2-3-sync-initial` left unchanged to avoid orphaning this PR's review history.

PR #106, opened 2026-05-24
M5.5: PasswordHasher (argon2-cffi)

### LINKED ISSUE #87: M5.5 PasswordHasher (argon2-cffi)
A small wrapper module over `argon2.PasswordHasher` (from `argon2-cffi`, added as a project dependency in M5.2). The wrapper folds `argon2-cffi`'s exception types into a boolean return for `verify`, so callers do not branch on `VerifyMismatchError` vs `InvalidHashError`. Cost parameters are constructor args so settings-driven tuning is a constructor call, not module-global state.

## File: `src/py/novamoc/domain/accounts/_password.py`

Shape:

```python
@dataclass(frozen=True, slots=True)
class PasswordHasher:
    time_cost: int = 3
    memory_cost_kib: int = 64 * 1024   # 64 MiB
    parallelism: int = 4

    @classmethod
    def from_defaults(cls) -> "PasswordHasher": ...

    def hash(self, password: str) -> str: ...
    def verify(self, encoded: str, password: str) -> bool: ...      # never raises
    def check_needs_rehash(self, encoded: str) -> bool: ...          # treat malformed hashes as needing rehash
```

Defaults align with OWASP / RFC 9106 for argon2id as of 2026. The login handler in M5.10 calls `check_needs_rehash` after every successful verify so active users upgrade for free on cost rotation.

The hasher is stateless and cheap to construct; the asgi factory in M5.11 builds one instance at app startup and stashes it on `app.state.password_hasher` for DI.

## Acceptance criteria

`uv run pytest tests/accounts/test_password.py` passes, covering:
- Hash-then-verify round-trip succeeds.
- Verify against the wrong password returns `False`, no exception.
- Two hashes of the same password differ (distinct salts).
- `check_needs_rehash` returns `True` after a cost-parameter bump on the verifier.
- Verify against a malformed encoded string returns `False`, no exception.

**Depends on.** #84 (M5.2 — adds `argon2-cffi` to `pyproject.toml`).

shoriminimoe:
## Summary

- Adds `src/py/novamoc/domain/accounts/_password.py`: a frozen dataclass wrapping `argon2.PasswordHasher` that folds all argon2-cffi exception types into a boolean return for `verify` (never raises), and treats malformed hashes as needing rehash in `check_needs_rehash`.
- Cost parameters (`time_cost`, `memory_cost_kib`, `parallelism`) are constructor args with OWASP/RFC 9106 defaults (m=64 MiB, t=3, p=4) so settings-driven tuning is a constructor call, not module-global state.
- Inner `argon2.PasswordHasher` is built lazily per call — cheap to construct, avoids `object.__setattr__` on a frozen dataclass.

Closes #87.

## Test plan

- [x] `uv run pytest tests/accounts/test_password.py -v` — 6/6 pass
- [x] `uv run ruff check src/py/novamoc/domain/accounts/_password.py tests/accounts/test_password.py` — clean
- [x] `uv run ty check src/py/novamoc/domain/accounts/_password.py` — 0 diagnostics
- [x] `just ratchet` — `Ratchet OK: 22 violations, baseline matches` (no new violations)
- [x] Hash-then-verify round-trip succeeds
- [x] Wrong password returns `False`, no exception
- [x] Two hashes of the same password differ (distinct salts)
- [x] `check_needs_rehash` returns `True` after cost-parameter bump on verifier
- [x] `verify` against malformed encoded string returns `False`, no exception
- [x] `check_needs_rehash` against malformed encoded string returns `True`, no exception

### DIFF
--- a/src/py/novamoc/domain/accounts/_password.py
+++ b/src/py/novamoc/domain/accounts/_password.py
@@ -19,6 +19,12 @@ from dataclasses import dataclass
 import argon2
 import argon2.exceptions
 
+# ``InvalidHashError`` is a ``ValueError`` rather than an ``Argon2Error``,
+# so both bases must be listed to fold every verify failure into ``False``.
+# Named so the ``except`` clause stays unambiguously a tuple of exception
+# types under both ``except A, B:`` and ``except (A, B):`` parser forms.
+_VERIFY_ERRORS = (argon2.exceptions.Argon2Error, argon2.exceptions.InvalidHashError)
+
 
 @dataclass(frozen=True, slots=True)
 class PasswordHasher:
@@ -81,7 +87,7 @@ class PasswordHasher:
         """
         try:
             return self._inner().verify(encoded, password)
-        except (argon2.exceptions.Argon2Error, argon2.exceptions.InvalidHashError):
+        except _VERIFY_ERRORS:
             return False
 
     def check_needs_rehash(self, encoded: str) -> bool:
--- a/src/py/novamoc/domain/accounts/_password.py
+++ b/src/py/novamoc/domain/accounts/_password.py
@@ -1,0 +1,104 @@
+"""PasswordHasher wrapper over argon2-cffi (ADR-020, M5.5).
+
+Wraps :class:`argon2.PasswordHasher` to fold the library's exception hierarchy
+into a boolean return for :meth:`verify` and :meth:`check_needs_rehash`, so
+callers never branch on ``VerifyMismatchError`` vs ``InvalidHashError``.  Cost
+parameters live on the dataclass so settings-driven tuning is a constructor
+call, not module-global state.
+
+Inner :class:`argon2.PasswordHasher` instances are built lazily on each
+call — the underlying object is cheap to construct (parameter validation only,
+no I/O) and a frozen dataclass cannot hold a mutable private attribute without
+``object.__setattr__``.  Lazy construction keeps the class simple.
+"""
+
+from __future__ import annotations
+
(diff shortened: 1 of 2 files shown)


shoriminimoe:
### Code review

Found 1 issue:

1. `SyntaxError` — Python 2-style `except A, B:` instead of `except (A, B):` means the module cannot be imported, so the test suite cannot have run. The PR description's "6/6 pass" / "ruff clean" / "ty 0 diagnostics" claims are not reproducible against this diff. Fix to `except (argon2.exceptions.Argon2Error, argon2.exceptions.InvalidHashError):` — both names are needed because `InvalidHashError` is a `ValueError`, not an `Argon2Error`.

https://github.com/shoriminimoe/novamoc/blob/6acbb0463f41bf702dd0130bf141f073222e0377/src/py/novamoc/domain/accounts/_password.py#L83-L85

shoriminimoe:
Verified on the repo's Python (3.14.3): the file parses, imports, and the test suite is in fact green against the original — Python 3.14's parser accepts the comma form and treats it as a tuple at the AST level (`ast.dump` of `except A, B:` shows `ExceptHandler(type=Tuple(elts=[Name('A'), Name('B')]))`). So the "can't have run" inference doesn't hold here.

That said, the [Python 3 docs](https://docs.python.org/3/reference/compound_stmts.html#the-try-statement) explicitly require the parenthesised form for multiple exception types, and the parens remove the visual ambiguity with `except A as B:`. Applied in 55f3bc6.

Issue #88, opened 2026-05-15
M5.6 AuthSettings + LoginFailedError + UserAlreadyHasTenantError as DomainErrors


shoriminimoe:
Three things, in one commit, because the error codes and the settings dataclass land in the same wiring pass:

## `AuthSettings` in `src/py/novamoc/config.py`

A new `@dataclass(frozen=True, slots=True)` matching the shape of `DatabaseSettings` / `ServerSettings` / `AppSettings`. Fields:

| Field | Env var | Default |
|---|---|---|
| `session_ttl_seconds: int` | `NOVAMOC_AUTH_SESSION_TTL_SECONDS` | `86400` (24h absolute TTL) |
| `session_cookie_name: str` | `NOVAMOC_AUTH_SESSION_COOKIE_NAME` | `"novamoc_session"` |
| `session_cookie_secure: bool` | `NOVAMOC_AUTH_SESSION_COOKIE_SECURE` | `False` (set to `true` in production) |
| `argon2_time_cost: int` | `NOVAMOC_AUTH_ARGON2_TIME_COST` | `3` |
| `argon2_memory_cost_kib: int` | `NOVAMOC_AUTH_ARGON2_MEMORY_COST_KIB` | `65536` (64 MiB) |
| `argon2_parallelism: int` | `NOVAMOC_AUTH_ARGON2_PARALLELISM` | `4` |

**Do not add `dev_seed_default_admin` or any other dev-mode flag** — the server has no environment-conditional code. Bootstrap is via CLI in every environment (M5.13 + M5.15).

Add `auth: AuthSettings = field(default_factory=AuthSettings)` to `Settings`.

## Error codes on `ErrorCode`

Append to `src/py/novamoc/domain/_errors.py::ErrorCode`:

```python
LOGIN_FAILED = "login_failed"
USER_ALREADY_HAS_TENANT = "user_already_has_tenant"
```

## Promote `UserAlreadyHasTenantError` to a `DomainError`; add `LoginFailedError`

`src/py/novamoc/domain/accounts/_errors.py`. Replace the bare-Exception `UserAlreadyHasTenantError` landed by M5.4 with a `DomainError` subclass (match the shape of subclasses under `domain/schema/_errors.py`):

```python
class LoginFailedError(DomainError):
    code = ErrorCode.LOGIN_FAILED
    status_code = 401
    default_message = "The provided credentials were not accepted."
    # Anti-enumeration: detail deliberately does not mention "password"
[...]

Issue #86, opened 2026-05-15
M5.4 UserTenantMembership model + service with N:1 write-time invariant


shoriminimoe:
Add the user ↔ tenant join table and the service-layer write check that enforces the v1 "one tenant per user" invariant.

## Model: `src/py/novamoc/db/models/_auth/_membership.py`

```python
class UserTenantMembership(DefaultBase):
    __tablename__ = "user_tenant_memberships"

    user_id: Mapped[uuid.UUID] = mapped_column(
        ForeignKey("users.id"), primary_key=True
    )
    tenant_id: Mapped[uuid.UUID] = mapped_column(
        ForeignKey("tenants.id"), primary_key=True
    )
```

- Composite PK `(user_id, tenant_id)` doubles as the uniqueness constraint.
- `DefaultBase` (not `UUIDAuditBase`) — the membership is a relation, not an entity; its identity is the pair, not an opaque id. No audit columns.
- The schema allows N-to-N from day one so v2's switch-tenant feature is a service-layer relaxation, not a schema migration.

Re-export from `src/py/novamoc/db/models/_auth/__init__.py`.

## `UserAlreadyHasTenantError` (interim bare Exception)

Append to `src/py/novamoc/domain/accounts/_errors.py`:

```python
class UserAlreadyHasTenantError(Exception):
    """Raised by UserTenantMembershipService.create when the user already
    has a membership. Promoted to a DomainError subclass with the proper
    ErrorCode plumbing in M5.6 (#88) — landing here as a bare Exception
    so M5.4 can ship without depending on the problem-details work.
    """
```

This handoff lets M5.4 land independently. M5.6 swaps the class to a `DomainError` subclass; the type identity persists so callers (and the M5.13 CLI) keep working through the swap.

## `UserTenantMembershipService`

Append to `src/py/novamoc/domain/accounts/_services.py`:

- `list_for_user(user_id: uuid.UUID) -> list[UserTenantMembership]`.
- `get_for_user(user_id: uuid.UUID) -> UserTenantMembership | None` — the invariant guarantees ≤1 row so callers (login handler in M5.10, CLI in M5.13) do not need to count.
- Override `create`: if `data["user_id"]` already has any membership (`list_for_user(user_id)` returns a non-empty list), raise `UserAlreadyHasTenantError` before the INSERT.

## Acceptance criteria

[...]

PR #111, opened 2026-05-24
feat(accounts): UserTenantMembership model + service (M5.4, #86)

shoriminimoe:
## Summary

Implements [#86](https://github.com/shoriminimoe/novamoc/issues/86) — adds the user ↔ tenant join table and the service-layer write check that enforces the v1 "one tenant per user" invariant from ADR-020.

- **`UserTenantMembership`** (`db/models/_auth/_membership.py`): composite PK `(user_id, tenant_id)`, ``DefaultBase`` (no audit columns, the pair is the identity). Schema is N-to-N so v2's switch-tenant feature is a service-layer relaxation, not a migration.
- **`UserAlreadyHasTenantError`** (interim bare `Exception`): documented to be promoted to a `DomainError` subclass in M5.6 (#88); type identity persists across the swap so callers (and the M5.13 CLI) keep working.
- **`UserTenantMembershipService`** (`domain/accounts/_services.py`): `list_for_user`, `get_for_user`, plus an overridden `create` that rejects a second membership for an existing `user_id`.
- **Tenant-scoping listener fix** (`db/_listeners.py`): the existing `"tenant_id in columns"` heuristic would have mis-classified the membership table — its `tenant_id` is a FK into the registry, not a scope predicate. Extended `_is_tenant_scoped` to honor an `info={"registry_fk": True}` opt-out marker on the column. CLAUDE.md updated to document the marker.

## Test plan

- [x] `uv run pytest tests/accounts/test_membership_model.py` — 5 tests: insert round-trip, composite-PK uniqueness, orphan-user FK rejection, orphan-tenant FK rejection, schema-allows-N-to-N.
- [x] `uv run pytest tests/accounts/test_membership_service.py` — 7 tests pinning the 1:1 invariant: first membership succeeds, second raises `UserAlreadyHasTenantError`, delete-then-recreate succeeds, `list_for_user` / `get_for_user` empty + populated.
- [x] Full suite: **456 passed**. `just ratchet` matches baseline (22 pre-existing violations).
- [x] `just typecheck` clean.

### One factual note

The issue's parenthetical claims "aiosqlite enables `PRAGMA foreign_keys=ON` by default" — it doesn't (verified). The model tests issue `PRAGMA foreign_keys = ON` per-connection so the orphan FK assertions actually fire. Production engine configuration is out of scope here; ADR-004 specifies it should be on, can be addressed in a follow-up.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -21,11 +21,11 @@ The schema-change log is **command-grain** (one row per accepted `POST /schema`)
   - `asgi.py` — `create_app()` factory used by Litestar/Granian.
   - `db/models/schema/` — server-authoritative meta-schema tables + `schema_change_log`.
   - `db/models/data/` — synced entity projections, `*_field_values` LWW projections, `event_log`.
-  - `db/models/_auth/` — auth / tenant-registry tables (`Tenant`; users + memberships land later in M5). Not tenant-scoped — rows in these tables *are* the tenants and users.
+  - `db/models/_auth/` — auth / tenant-registry tables (`Tenant`, `User`, `Session`, `UserTenantMembership`). Not tenant-scoped — rows in these tables *are* the tenants and users; the membership's `tenant_id` opts out of the scoping listeners via `info={"registry_fk": True}` because it points at the registry rather than declaring scope.
   - `db/models/_mixins.py` — `TenantScopedMixin` (adds `tenant_id: Mapped[uuid.UUID]` PK column via `GUID` with `sort_order=-200` so the composite PK leads with `tenant_id`, ADR-014/ADR-020).
   - `db/_listeners.py` — three SQLAlchemy event listeners that enforce tenant scoping (issue #51); `db/_tenant_context.py` holds the request-scoped `current_tenant_id` ContextVar and `use_tenant` helper.
   - `domain/schema/` — `POST /schema` endpoint stack (commands, payloads, dispatch, handlers, services, controller). See "Schema endpoint" below.
-  - `domain/accounts/` — authentication middleware + tenant resolver + `TenantService` (advanced-alchemy service for the `tenants` registry).
+  - `domain/accounts/` — authentication middleware + tenant resolver + registry services (`TenantService`, `UserService`, `UserTenantMembershipService`). The membership service enforces the v1 one-membership-per-user invariant at write time via `UserAlreadyHasTenantError` (ADR-020).
 - `src/js/web/` — Svelte 5 + Vite + Tailwind SPA (currently scaffolding only).
 - `tests/` — pytest suite (currently schema endpoint only).
 - `docs/adr/` — accepted/proposed architecture decisions (ADR-000 through ADR-016).
@@ -104,7 +104,7 @@ Cross-tenant isolation is enforced structurally by three SQLAlchemy event listen
 
 Handler call sites consequently pass NO `tenant_id` for reads and creates (auto-handled). They DO pass it inside `item_id=(auth.tenant_id, req.entity_id)` for `update`/`delete` because the composite PK is the WHERE clause. The `SchemaChangeLogService.append`/`current_version` API takes no `tenant_id` arg — Layer 1 supplies the predicate.
 
-Escape hatch: the `SKIP_TENANT_FILTER` execution option suppresses Layer 1 and Layer 3. Greppable, no production callers in v1.
+Escape hatch: the `SKIP_TENANT_FILTER` execution option suppresses Layer 1 and Layer 3. Greppable, no production callers in v1. A separate marker — `info={"registry_fk": True}` on a `tenant_id` column — flags it as a FK into the `tenants` registry rather than a scope predicate, so the listeners ignore the table entirely (used by `user_tenant_memberships`).
 
 Out of scope of these layers: raw `session.execute(text(...))` SELECTs and in-place `obj.tenant_id` mutation. No production code uses these patterns; if added in the future, audit them carefully — they bypass the structural enforcement.
 
--- a/src/py/novamoc/db/_listeners.py
+++ b/src/py/novamoc/db/_listeners.py
@@ -31,8 +31,15 @@ from novamoc.db._errors import CrossTenantWriteError, UnscopedQueryError
 from novamoc.db._tenant_context import SKIP_TENANT_FILTER, current_tenant_id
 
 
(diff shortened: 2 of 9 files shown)


shoriminimoe (commented):
Review from extra-high-recall pass (8 findings, ranked by severity).

## HIGH — invariant bypasses in `_extract_user_id`

The v1 one-membership-per-user invariant has two distinct escape hatches.

### 1. Non-dict / non-model payloads silently bypass the check
`src/py/novamoc/domain/accounts/_services.py:113-123`

`_extract_user_id` only recognises `dict` and `UserTenantMembership` instances, but the signature accepts `ModelDictT[UserTenantMembership]`, which per advanced_alchemy's typing module (`service/typing.py:83`) admits `msgspec.Struct`, `pydantic.BaseModel`, attrs classes, and `DTOData`. For any of those shapes the function falls through to `return None`, the `if user_id is not None` guard at line 108 short-circuits, and `super().create()` inserts a row — no `UserAlreadyHasTenantError` ever raised.

The docstring at line 116 ("regardless of shape") is precisely the trap: a future caller passing a Struct or pydantic model will trust the invariant is enforced.

### 2. String-form UUIDs in dict payload also bypass the check
`src/py/novamoc/domain/accounts/_services.py:121`

`isinstance(value, uuid.UUID)` rejects string UUIDs, but `GUID.process_bind_param` (`advanced_alchemy/types/guid.py:53`) accepts `str | bytes | UUID`, so the INSERT succeeds. A caller passing `{"user_id": "550e8400-...", "tenant_id": "..."}` (routine from JSON or argparse) bypasses the check silently.

**Suggested fix:** either accept string UUIDs (`try: return uuid.UUID(value)`) plus a `getattr(data, "user_id", None)` fallback for Struct/pydantic, OR drop `_extract_user_id` entirely and add a DB-level `UniqueConstraint("user_id")` for v1, removing it in v2. The DB-level approach also closes finding 5.

## MEDIUM — missing FK cascade behaviour
`src/py/novamoc/db/models/_auth/_membership.py:28,37`

Neither `ForeignKey("users.id")` nor `ForeignKey("tenants.id")` declares `ondelete=`. SQLite defaults to NO ACTION. With PRAGMA foreign_keys=ON (the membership tests' setup and ADR-004's production intent), deleting a user/tenant with a live membership raises IntegrityError; with PRAGMA OFF (current production default per the PR's own factual note) it silently orphans. The cascade intent for v1 is undeclared — likely `CASCADE` on user delete, `RESTRICT` on tenant delete, but it should be explicit.

## MEDIUM — TOCTOU race between check and insert
`src/py/novamoc/domain/accounts/_services.py:108-110`

SELECT-then-INSERT on independent sessions with no row lock, no SAVEPOINT, no `user_id`-alone UNIQUE constraint. Two concurrent calls with the same `user_id` and different `tenant_ids` both pass the check and both INSERT (composite PK doesn't collide on different tenant_ids). No production callers exist yet, but the docstring's "enforces the v1 invariant at write time" overpromises. A DB-level UNIQUE on `user_id` is the clean fix and would also remove the need for `_extract_user_id`.

## LOW — downstream + coverage

### `get_for_user` raises MultipleResultsFound if the invariant is violated
`src/py/novamoc/domain/accounts/_services.py:87-94`

`get_one_or_none` raises if more than one row matches. Combined with the bypasses above, this propagates as an uncaught 500 and locks callers out of even reading the user's primary tenant to repair the state. Consider `get_first()` or `list_for_user(...)[:1]` if degraded-but-readable is the desired posture; or rely on the DB-level UNIQUE so this can't happen.

### `registry_fk` opt-out is not exercised by listener tests
`src/py/novamoc/db/_listeners.py:40-42`

No test in `tests/listeners/test_layer{1,2,3}*.py` constructs a synthetic table with `info={"registry_fk": True}`. The membership tests (`@pytest.mark.no_tenant`) don't actually drive the listeners. A future refactor of any layer that re-derives the column check locally would regress the opt-out silently. One synthetic-table fixture per layer test would lock this in.

### Docstring overpromise
`src/py/novamoc/domain/accounts/_services.py:116`

"Pull `user_id` out of a create payload regardless of shape" — currently only two shapes are handled. Either match the promise or scope the docstring.

---

Nothing here is a release-blocker on its own given v1 has no callers of the service yet, but findings 1 and 2 are the kind of silent escape hatch that's easy to ship and hard to notice once exercised, and finding 3 (missing FK cascade) will bite the first time someone tries to delete a user or tenant in a test or admin script. The DB-level `UniqueConstraint("user_id")` + explicit `ondelete=` would close 1, 2, 3, 5, and 7 in one move.

shoriminimoe:
Thanks for the thorough review. Pushed [`8e067f9`](https://github.com/shoriminimoe/novamoc/commit/8e067f9) addressing all eight findings:

| # | Severity | Resolution |
|---|---|---|
| 1 | HIGH | `_extract_user_id` now uses `getattr(data, "user_id", None)` fallback for any non-dict/non-ORM shape (Structs, pydantic, attrs, dataclasses). |
| 2 | HIGH | String UUIDs in dict payloads are parsed (`uuid.UUID(value)`). |
| 1+2+4 (backstop) | HIGH/MED | Added `UniqueConstraint("user_id", name="uq_user_tenant_memberships_user_id")` as the structural promise. Any bypass — including TOCTOU — now raises `IntegrityError` at the DB. The service docstring documents this as the backstop; callers needing the friendly error for exotic shapes catch IntegrityError themselves. v2 drops the constraint alongside the service-layer check; pre-release breaking changes are fine per CLAUDE.md. |
| 3 | MED | Explicit `ondelete="CASCADE"` on user FK, `ondelete="RESTRICT"` on tenant FK. Both fire only under `PRAGMA foreign_keys=ON`; ADR-004's production wiring remains a follow-up. |
| 5 | LOW | With `UNIQUE(user_id)`, `get_one_or_none` in `get_for_user` is structurally safe — `MultipleResultsFound` is impossible. |
| 6 | LOW | New `tests/db/test_registry_fk_marker.py` covers the `info={"registry_fk": True}` opt-out across all three listener layers. Uses the real membership table (the only and canonical caller); a future refactor regressing the marker check now breaks layer-focused tests. |
| 7 | LOW | Rewrote the `_extract_user_id` docstring to match what it actually does. |

New tests: 5 (model — unique, cascade, restrict, distinct-users; service — string-UUID, DB-backstop) + 5 (listener marker). Full suite **466 passed**, ratchet matches baseline.

The model docstring also now reflects the design change: schema enforces 1:1 alongside the service; v2 drops both together.

PR #115, opened 2026-05-25
feat(cli): novamoc tenant / user / auth commands (M5.13, #94)

### LINKED ISSUE #94: M5.13 novamoc CLI: tenant / user / auth commands
Click-based operator CLI exposed as the `novamoc` console script via `[project.scripts]` in `pyproject.toml`. The CLI is the **only** bootstrap path — production deployments run it in an init container; local dev runs it via `just bootstrap-dev` (M5.15).

## Entry point

Add to `pyproject.toml`:

```toml
[project.scripts]
novamoc = "novamoc.cli:main"
```

Run `uv sync` to wire the script.

## File: `src/py/novamoc/cli.py`

A `@click.group()` named `main` with three sub-groups:

### `novamoc tenant create --display-name <name>`

Creates a tenant. PK UUID is generated by the DB (via `UUIDAuditBase`'s default factory). On success prints `Created tenant <uuid>.` so the recipe (M5.15) can capture the UUID with `awk`. Exits non-zero with a human-readable error to stderr on failure.

### `novamoc user create <username> [--password <pw>]`

Creates a user. If `--password` is omitted, prompt with `click.prompt(..., hide_input=True)` for interactive use; tests pass `--password` to skip the prompt. Hashes the password via `PasswordHasher` constructed from `AuthSettings`. Exits non-zero on uniqueness violation with `User '<x>' already exists.`

### `novamoc user set-password <username> [--password <pw>]`

Operator-driven password reset (no email flow in v1). Same `--password` semantics as `user create`. Errors `User '<x>' not found.` if no matching row.

### `novamoc user add-to-tenant <username> <tenant-uuid>`

Adds the membership. The `<tenant-uuid>` arg is parsed as a `uuid.UUID`; invalid syntax exits with `Error: '<x>' is not a valid UUID.` Username lookup uses case-folded match (same fold as `UserService.create`).

If the membership service raises `UserAlreadyHasTenantError`, exit non-zero with stderr message including the substring "`already`" and "`tenant`" (so the test in this issue can grep for both). The wire-level mapping of `UserAlreadyHasTenantError` to 409 problem-details lives on the HTTP surface; the CLI uses the same exception to produce its exit code.

### `novamoc user exists <username>`

Exit code 0 if the user exists, 1 if not. No stdout output (just the exit code). Used by `just bootstrap-dev` (M5.15) to short-circuit the bootstrap if `admin` is already there.

### `novamoc auth gc-sessions`

Deletes `Session` rows where `expires_at < now(UTC)`. Prints `Deleted N expired sessions.` so the operator can confirm. The opportunistic GC the design records as tech debt; future scheduler work makes this automatic.

## Session lifecycle inside the CLI

Each sub-command opens its own `AsyncSession` against the configured DB (via `Settings()` reading env vars), commits on success, rolls back and exits non-zero on any exception. Uses `asyncio.run(...)` at the entry point per Click's idiom.

## Tests: `tests/accounts/test_cli.py`

Use `click.testing.CliRunner`. Tests need a real database — point `NOVAMOC_DB_URL` at a `tmp_path` SQLite file via fixture, or call `Settings()` with overrides.

Coverage:
- `tenant create` succeeds; output contains a parseable UUID.
- `user create alice --password x` succeeds.
- `user set-password alice --password y` succeeds.
- `user exists alice` exits 0 after create; `user exists ghost` exits 1.
- `user add-to-tenant alice <uuid>` succeeds for the first tenant.
- `user add-to-tenant alice <second-uuid>` exits non-zero, stderr matches `/already.*tenant/i` (the N:1 rejection — the CLI-side test of the M5.4 invariant).
- `auth gc-sessions` runs on an empty `sessions` table without error and prints a deletion count.

## Acceptance criteria

- `uv run pytest tests/accounts/test_cli.py` passes.
- `uv run novamoc --help` prints the command tree.
- `just ratchet` clean.

**Depends on.** #86 (M5.4 — membership service raises the N:1 error), #87 (M5.5 — `PasswordHasher` for create/set-password), #88 (M5.6 — `UserAlreadyHasTenantError` as `DomainError`).

shoriminimoe:
Closes #94.

## Summary

Adds the Click-based ``novamoc`` operator CLI as the sole bootstrap path for tenants, users, and memberships. Production deployments run it in an init container; local dev will wire it through ``just bootstrap-dev`` (M5.15, follow-up).

Each sub-command opens its own ``AsyncSession`` against the engine built from ``Settings()`` (so ``NOVAMOC_DB_URL`` and the argon2 cost parameters are read at invocation time), commits on success, and exits non-zero with a human-readable stderr message on failure. The CLI runs outside the request lifecycle; the auth-registry tables are not tenant-scoped, so the listeners short-circuit naturally on column absence.

## Commands

- ``novamoc tenant create --display-name <name>`` — prints ``Created tenant <uuid>.`` so M5.15 can ``awk`` the new id out.
- ``novamoc user create <username> [--password <pw>]`` — hashes via ``PasswordHasher`` from ``AuthSettings``; pre-checks for case-folded duplicates so the operator gets a friendly ``User '<x>' already exists.`` before the ``UNIQUE`` backstop fires. Without ``--password``, prompts interactively (``hide_input=True``, confirmation).
- ``novamoc user set-password <username> [--password <pw>]`` — operator-driven reset (no email flow in v1). New hash uses current cost parameters.
- ``novamoc user exists <username>`` — exit 0 / 1 only, no stdout output. Used to short-circuit the M5.15 bootstrap when the seed user is already present.
- ``novamoc user add-to-tenant <username> <tenant-uuid>`` — username lookup is case-folded via ``UserService.get_by_username``. Surfaces ``UserAlreadyHasTenantError`` from M5.4 as a CLI failure containing both "already" and "tenant" so the bootstrap recipe can grep for the N:1 case without coupling to the full sentence. Invalid UUID syntax exits with ``'<x>' is not a valid UUID.``
- ``novamoc auth gc-sessions`` — opportunistic GC of expired ``sessions`` rows; prints ``Deleted N expired sessions.`` A future scheduler makes this automatic.

## Non-goals

- ``just bootstrap-dev`` is M5.15, not in this PR.
- Wire-level mapping of ``UserAlreadyHasTenantError`` to 409 problem-details lives on the HTTP surface; the CLI uses the same exception only for its exit code.
- No historical / as-of-seq operations.

## Test plan

- [x] ``uv run pytest tests/accounts/test_cli.py`` — 13 tests covering: ``--help`` lists groups; ``tenant create`` prints a parseable UUID; ``user create`` succeeds and rejects duplicates; ``set-password`` succeeds and rejects missing users; ``user exists`` returns 0/1 with no stdout; ``add-to-tenant`` succeeds, rejects invalid UUIDs, rejects missing users, and rejects the second tenant for an already-mapped user (the M5.4 invariant test on the CLI side); ``auth gc-sessions`` runs cleanly on an empty table.
- [x] ``uv run novamoc --help`` prints the command tree.
- [x] ``just lint``, ``just typecheck``, ``just test-py`` all clean.
- [x] ``just ratchet`` clean (coverage baseline bumped to reflect the new CLI tests).

### DIFF
--- a/src/py/novamoc/cli.py
+++ b/src/py/novamoc/cli.py
@@ -31,6 +31,7 @@ from typing import TYPE_CHECKING
 
 import click
 from sqlalchemy import delete
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.asyncio import async_sessionmaker, create_async_engine
 
 # Register tenant-scoping event handlers on SQLAlchemy. Today's CLI
@@ -200,22 +201,29 @@ def user_create(username: str, password: str | None) -> None:
     )
     settings = _load_settings()
     hasher = _password_hasher(settings)
-    hashed = hasher.hash(plaintext)
 
     async def work(session: AsyncSession) -> None:
         svc = UserService(session=session)
-        # Pre-check: the service folds usernames at create-time, so a
-        # case-folded lookup catches duplicates before SQLAlchemy raises
-        # IntegrityError. The UNIQUE constraint on ``users.username``
-        # remains the structural backstop.
+        # Pre-check before hashing: the service folds usernames at
+        # create-time, so a case-folded lookup catches duplicates
+        # before the ~100 ms argon2id cost. The UNIQUE constraint on
+        # ``users.username`` remains the structural backstop.
         existing = await svc.get_by_username(username)
         if existing is not None:
             msg = f"User '{username}' already exists."
             raise click.ClickException(msg)
-        await svc.create(
-            data={"username": username, "password_hash": hashed},
-            auto_commit=False,
-        )
+        hashed = hasher.hash(plaintext)
+        try:
+            await svc.create(
+                data={"username": username, "password_hash": hashed},
+                auto_commit=False,
+            )
+        except IntegrityError as exc:
+            # Concurrent CLI invocations can race past the pre-check;
+            # the UNIQUE backstop's IntegrityError gets the same
+            # friendly message rather than a raw SQLAlchemy traceback.
+            msg = f"User '{username}' already exists."
+            raise click.ClickException(msg) from exc
 
     asyncio.run(_run_in_session(settings, work))
     click.echo(f"Created user {username}.")
@@ -239,15 +247,16 @@ def user_set_password(username: str, password: str | None) -> None:
     )
     settings = _load_settings()
     hasher = _password_hasher(settings)
-    hashed = hasher.hash(plaintext)
 
     async def work(session: AsyncSession) -> None:
         svc = UserService(session=session)
+        # Pre-check before hashing: lookup is far cheaper than the
+        # ~100 ms argon2id work for a username that doesn't exist.
         existing = await svc.get_by_username(username)
         if existing is None:
             msg = f"User '{username}' not found."
             raise click.ClickException(msg)
-        existing.password_hash = hashed
+        existing.password_hash = hasher.hash(plaintext)
         # No explicit flush — ``_run_in_session`` commits, which
         # flushes implicitly.
 
@@ -311,6 +320,12 @@ def user_add_to_tenant(username: str, tenant_uuid: str) -> None:
             # both "already" and "tenant" in the message lets the M5.15
             # bootstrap recipe (and the test) grep for the N:1 case
             # without coupling to the full sentence.
+            msg = f"User '{username}' already belongs to a tenant."
+            raise click.ClickException(msg) from exc
+        except IntegrityError as exc:
+            # Concurrent CLI invocations can race past the service-layer
+            # pre-check; the UNIQUE(user_id) backstop's IntegrityError
+            # gets the same friendly message.
             msg = f"User '{username}' already belongs to a tenant."
             raise click.ClickException(msg) from exc
 
--- a/src/py/novamoc/cli.py
+++ b/src/py/novamoc/cli.py
@@ -55,6 +55,20 @@ if TYPE_CHECKING:
     from sqlalchemy.ext.asyncio import AsyncSession
 
 
+def _load_settings() -> Settings:
+    """Construct ``Settings`` with parse errors mapped to ``ClickException``.
+
+    Env-var parsing (``_int_env`` etc.) raises ``ValueError`` on a bad
+    value; surfacing that as a raw traceback contradicts the module
+    docstring's promise of a friendly stderr message. The mapping is
+    central so every command that loads settings benefits.
+    """
+    try:
+        return Settings()
+    except ValueError as exc:
+        raise click.ClickException(str(exc)) from exc
+
+
 def _password_hasher(settings: Settings) -> PasswordHasher:
     """Build a ``PasswordHasher`` from the configured auth parameters."""
     return PasswordHasher(
@@ -132,7 +146,7 @@ def tenant() -> None:
 )
 def tenant_create(display_name: str) -> None:
     """Create a tenant and print its UUID."""
-    settings = Settings()
+    settings = _load_settings()
 
     async def work(session: AsyncSession) -> uuid.UUID:
         # Return the UUID directly rather than handing back an ORM
@@ -166,7 +180,7 @@ def user() -> None:
 )
 def user_create(username: str, password: str | None) -> None:
     """Create a user with a hashed password."""
-    settings = Settings()
+    settings = _load_settings()
     hasher = _password_hasher(settings)
     plaintext = _prompt_password_if_missing(password)
     hashed = hasher.hash(plaintext)
@@ -203,7 +217,7 @@ def user_set_password(username: str, password: str | None) -> None:
     Operator-driven: no email-confirmation flow in v1. The new hash
     uses the current cost parameters from :class:`AuthSettings`.
     """
-    settings = Settings()
+    settings = _load_settings()
     hasher = _password_hasher(settings)
     plaintext = _prompt_password_if_missing(password)
     hashed = hasher.hash(plaintext)
@@ -231,7 +245,7 @@ def user_exists(username: str) -> None:
     by ``just bootstrap-dev`` (M5.15) to skip the seed when the
     target user is already provisioned.
     """
-    settings = Settings()
+    settings = _load_settings()
 
     async def work(session: AsyncSession) -> bool:
         svc = UserService(session=session)
@@ -259,7 +273,7 @@ def user_add_to_tenant(username: str, tenant_uuid: str) -> None:
         msg = f"'{tenant_uuid}' is not a valid UUID."
         raise click.ClickException(msg) from None
 
-    settings = Settings()
+    settings = _load_settings()
 
     async def work(session: AsyncSession) -> None:
         users = UserService(session=session)
@@ -303,7 +317,7 @@ def auth_gc_sessions() -> None:
     debt (ADR-020). Operators invoke this on a cadence appropriate
     to their session TTL.
     """
-    settings = Settings()
+    settings = _load_settings()
 
     async def work(session: AsyncSession) -> int:
         now = datetime.now(UTC)
--- a/src/py/novamoc/cli.py
+++ b/src/py/novamoc/cli.py
@@ -1,0 +1,315 @@
+"""Operator CLI exposed as the ``novamoc`` console script (M5.13).
+
+The CLI is the only bootstrap path for tenants, users, and
+memberships — production deployments run it in an init container,
+local dev wires it through ``just bootstrap-dev`` (M5.15). Each
+sub-command builds its own async engine + session from
+``Settings()`` (so it picks up ``NOVAMOC_DB_URL`` and the auth
+cost parameters at invocation time), runs the operation,
+commits on success, rolls back and exits non-zero with a
+human-readable stderr message on any failure.
+
+Output convention: success messages on stdout, error messages on
+stderr. The ``tenant create`` success line ends with the new
+tenant's UUID so the M5.15 recipe can ``awk`` it out.
+
+The CLI runs outside the request lifecycle so it does not use the
+tenant-scoping middleware or contextvar. The auth-registry tables
+are not tenant-scoped (they have no ``tenant_id`` column), so the
+listeners short-circuit naturally.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import sys
+import uuid
+from datetime import UTC, datetime
+from typing import TYPE_CHECKING
+
+import click
+from sqlalchemy import delete
+from sqlalchemy.ext.asyncio import async_sessionmaker, create_async_engine
+
+from novamoc.config import Settings
+from novamoc.db.models._auth import Session
+from novamoc.domain.accounts._errors import UserAlreadyHasTenantError
+from novamoc.domain.accounts._password import PasswordHasher
+from novamoc.domain.accounts._services import (
+    TenantService,
+    UserService,
+    UserTenantMembershipService,
+)
+
+if TYPE_CHECKING:
+    from collections.abc import Awaitable, Callable
+
+    from sqlalchemy.ext.asyncio import AsyncSession
+
+    from novamoc.db.models._auth import Tenant
+
+
+def _password_hasher(settings: Settings) -> PasswordHasher:
+    """Build a ``PasswordHasher`` from the configured auth parameters."""
+    return PasswordHasher(
+        time_cost=settings.auth.argon2_time_cost,
+        memory_cost_kib=settings.auth.argon2_memory_cost_kib,
+        parallelism=settings.auth.argon2_parallelism,
+    )
+
+
+async def _run_in_session[T](
+    settings: Settings,
+    work: Callable[[AsyncSession], Awaitable[T]],
+) -> T:
+    """Run ``work`` inside a fresh ``AsyncSession``.
+
+    Commits on success, rolls back on any exception (which then
+    propagates to the caller for the click error mapping). The
+    engine is disposed before returning so the CLI process exits
+    cleanly.
+    """
+    engine = create_async_engine(settings.db.url)
+    try:
+        factory = async_sessionmaker(engine, expire_on_commit=False)
+        async with factory() as session:
+            try:
+                result = await work(session)
+            except BaseException:
+                await session.rollback()
+                raise
+            await session.commit()
+            return result
+    finally:
+        await engine.dispose()
+
+
+def _prompt_password_if_missing(password: str | None) -> str:
+    """Prompt with hidden input when ``--password`` was not supplied.
+
+    Confirmation is required so a typo at the operator console does
+    not silently set an unrecoverable password.
+    """
+    if password is not None:
+        return password
+    return click.prompt("Password", hide_input=True, confirmation_prompt=True)
+
+
+def _fail(message: str) -> None:
+    """Emit ``message`` on stderr and exit non-zero."""
+    raise click.ClickException(message)
+
+
+@click.group()
+def main() -> None:
+    """novaMOC operator CLI.
+
+    Sub-groups:
+
+    * ``tenant`` — tenant registry administration.
+    * ``user`` — user registry administration.
+    * ``auth`` — session / credential maintenance.
+    """
+
+
+# ---------------------------------------------------------------------------
+# tenant
+# ---------------------------------------------------------------------------
+
+
+@main.group()
+def tenant() -> None:
+    """Tenant registry administration."""
+
+
+@tenant.command("create")
+@click.option(
+    "--display-name", required=True, help="Human-readable name for the tenant."
+)
+def tenant_create(display_name: str) -> None:
+    """Create a tenant and print its UUID."""
+    settings = Settings()
+
+    async def work(session: AsyncSession) -> Tenant:
+        svc = TenantService(session=session)
+        return await svc.create(data={"display_name": display_name}, auto_commit=False)
+
+    created = asyncio.run(_run_in_session(settings, work))
+    click.echo(f"Created tenant {created.id}.")
+
+
+# ---------------------------------------------------------------------------
+# user
+# ---------------------------------------------------------------------------
+
+
+@main.group()
+def user() -> None:
+    """User registry administration."""
+
+
+@user.command("create")
+@click.argument("username")
+@click.option(
+    "--password",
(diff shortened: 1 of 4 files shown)


src/py/novamoc/cli.py
```diff
@@ -0,0 +1,315 @@
...
+    async def work(session: AsyncSession) -> None:
+        users = UserService(session=session)
+        target = await users.get_by_username(username)
+        if target is None:
+            msg = f"User '{username}' not found."
+            raise click.ClickException(msg)
+        memberships = UserTenantMembershipService(session=session)
+        try:
+            await memberships.create(
+                data={"user_id": target.id, "tenant_id": tenant_id},
```
shoriminimoe, line 337: **[1] `user add-to-tenant` creates a membership pointing at a non-existent tenant UUID.** PRAGMA `foreign_keys=ON` is never enabled — grep finds zero connect-event listeners wiring it, and `_membership.py:40-41` acknowledges 'production wiring is a separate follow-up'. Operator runs `novamoc user add-to-tenant alice 00000000-0000-0000-0000-000000000000` (typo'd UUID, valid format): UUID parses, user exists, INSERT succeeds against a phantom tenant_id. CLI prints `Added user alice to tenant 0000…` and the row dangles. `test_user_add_to_tenant_invalid_uuid_exits_nonzero` only covers parse failure. Add an explicit tenant lookup before the membership create, or wire PRAGMA at the engine-creation site.

src/py/novamoc/cli.py
```diff
@@ -0,0 +1,315 @@
...
+        await engine.dispose()
+
+
+def _prompt_password_if_missing(password: str | None) -> str:
+    """Prompt with hidden input when ``--password`` was not supplied.
+
+    Confirmation is required so a typo at the operator console does
+    not silently set an unrecoverable password.
+    """
+    if password is not None:
```
shoriminimoe, line 117: **[2] `--password ''` is accepted as a valid password.** The check is `is not None`, so the empty string skips the confirm-prompt path; argon2 happily hashes `''`; the user row commits with an empty-password hash and the account is login-able with no credential. Trivially triggered by `--password "$PW"` where `PW` is unset. Click's interactive `prompt(confirmation_prompt=True)` rejects empty input, but the `--password` flag does not. Same risk in `user set-password`.

src/py/novamoc/cli.py
```diff
@@ -0,0 +1,315 @@
...
+        if target is None:
+            msg = f"User '{username}' not found."
+            raise click.ClickException(msg)
+        memberships = UserTenantMembershipService(session=session)
+        try:
+            await memberships.create(
+                data={"user_id": target.id, "tenant_id": tenant_id},
+                auto_commit=False,
+            )
+        except UserAlreadyHasTenantError as exc:
```
shoriminimoe, line 340: **[3] Only `UserAlreadyHasTenantError` is caught; the `UNIQUE(user_id)` backstop's `IntegrityError` is the caller's responsibility per `_services.py:113-115` and is not handled.** Two concurrent CLI invocations (init-container retry, two operator terminals) both pass `list_for_user` before either commits; the loser raises `sqlalchemy.exc.IntegrityError` which propagates through `asyncio.run` → `main` as a raw SQLAlchemy traceback instead of the documented 'already belongs to a tenant.' message. `user_create` (line 175) has the same shape against `users.username` UNIQUE.

src/py/novamoc/cli.py
```diff
@@ -0,0 +1,315 @@
...
+tenant's UUID so the M5.15 recipe can ``awk`` it out.
+
+The CLI runs outside the request lifecycle so it does not use the
+tenant-scoping middleware or contextvar. The auth-registry tables
+are not tenant-scoped (they have no ``tenant_id`` column), so the
+listeners short-circuit naturally.
+"""
+
+from __future__ import annotations
+
```
shoriminimoe, line 25: **[4] cli.py does not `import novamoc.db._listeners`.** The tenant-scoping listeners register as an import-time side effect from `asgi.py:38` only. Today every CLI command touches non-tenant-scoped registry tables so this is harmless — but tests import the listeners via `tests/conftest.py:21`, so tests run with structural enforcement while production CLI runs without it. The day someone adds a CLI command touching `assets` or `asset_field_values` (e.g. `novamoc data dump --tenant-id …`), tests catch a missing tenant context with `UnscopedQueryError` while production silently returns/inserts cross-tenant rows. Mirror asgi.py's `import novamoc.db._listeners  # noqa: F401` here.

shoriminimoe (commented):
Code review (extra-high effort, 15 findings). Most severe first — finding 1 is a real production gap; 2–3 are missing input validation / error handling; 4–7 are robustness/test gaps; 8–15 are quality and consistency.

src/py/novamoc/cli.py
```diff
@@ -0,0 +1,315 @@
...
+    """Tenant registry administration."""
+
+
+@tenant.command("create")
+@click.option(
+    "--display-name", required=True, help="Human-readable name for the tenant."
+)
+def tenant_create(display_name: str) -> None:
+    """Create a tenant and print its UUID."""
+    settings = Settings()
```
shoriminimoe: **[5] `Settings()` parse errors propagate as uncaught traceback, contradicting the module docstring.** The docstring promises 'rolls back and exits non-zero with a human-readable stderr message on any failure', but `Settings()` is called outside `_run_in_session`. Operator misconfigures `NOVAMOC_AUTH_ARGON2_TIME_COST=three` (or any other env-backed setting): `_int_env` raises `ValueError("cannot parse 'three' as int for NOVAMOC_AUTH_ARGON2_TIME_COST")` and Click renders a full Python traceback. Wrap `Settings()` (or factor the call into a helper that translates `ValueError` → `ClickException`).

tests/accounts/test_cli.py
```diff
@@ -0,0 +1,203 @@
...
+    assert result.exit_code != 0
+    assert re.search(r"already.*tenant", result.stderr, re.IGNORECASE)
+
+
+def test_auth_gc_sessions_on_empty_table_prints_count(
+    runner: CliRunner, db_url: str
+) -> None:
+    result = runner.invoke(main, ["auth", "gc-sessions"])
+    assert result.exit_code == 0, (result.output, result.stderr)
+    assert "Deleted 0 expired sessions." in result.stdout
```
shoriminimoe, line 234: **[6] `auth gc-sessions` is only tested against an empty table.** The actual delete predicate `expires_at < now` is never exercised. A refactor that flips the comparison to `expires_at > now` (deleting non-expired sessions) or filters on the wrong column still produces `rowcount=0` against an empty table → this test still passes. Add a row each side of `now` and assert only the expired one is deleted.

src/py/novamoc/cli.py
```diff
@@ -0,0 +1,315 @@
...
+    """
+    settings = Settings()
+
+    async def work(session: AsyncSession) -> bool:
+        svc = UserService(session=session)
+        return (await svc.get_by_username(username)) is not None
+
+    found = asyncio.run(_run_in_session(settings, work))
+    if not found:
+        sys.exit(1)
```
shoriminimoe, line 298: **[7] `user exists` conflates 'user absent' (exit 1) with 'CLI errored' (also exit 1).** M5.15's `just bootstrap-dev` runs `novamoc user exists seed-user` and seeds when exit != 0. If `NOVAMOC_DB_URL` is wrong or the DB is unreachable, the SQLAlchemy error → Click standalone wrapper exits 1 → recipe interprets as 'not present' → falls through to `user create` which fails the same way. A misconfigured environment looks identical to a fresh one until the create attempt's traceback is read. Use exit 2 for errors and reserve 0/1 for the answer.

src/py/novamoc/cli.py
```diff
@@ -0,0 +1,315 @@
...
+
+
+@main.group()
+def tenant() -> None:
+    """Tenant registry administration."""
+
+
+@tenant.command("create")
+@click.option(
+    "--display-name", required=True, help="Human-readable name for the tenant."
```
shoriminimoe, line 160: **[8] `tenant create --display-name ''` is accepted.** Click's `required=True` only rejects absent, not empty; no service-layer validation either. Row commits with empty `display_name`, tenant is unidentifiable in admin listings, and v1 has no rename command. Add a non-empty check or use a custom `type=` validator.

src/py/novamoc/cli.py
```diff
@@ -0,0 +1,315 @@
...
+        if existing is not None:
+            msg = f"User '{username}' already exists."
+            raise click.ClickException(msg)
+        await svc.create(
+            data={"username": username, "password_hash": hashed},
+            auto_commit=False,
+        )
+
+    asyncio.run(_run_in_session(settings, work))
+    click.echo(f"Created user {username}.")
```
shoriminimoe: **[9] Echo prints the raw `username` argument, not the NFKC+casefold form actually persisted.** Operator runs `novamoc user create ALICE --password x`; `UserService.create` folds to 'alice' before INSERT; CLI prints 'Created user ALICE.' Operator greps for 'ALICE' and finds nothing. Same in `set-password:212` and `add-to-tenant:275`. Echo the folded form (or re-read the persisted row).

src/py/novamoc/cli.py
```diff
@@ -0,0 +1,315 @@
...
+    hashed = hasher.hash(plaintext)
+
+    async def work(session: AsyncSession) -> None:
+        svc = UserService(session=session)
+        existing = await svc.get_by_username(username)
+        if existing is None:
+            msg = f"User '{username}' not found."
+            raise click.ClickException(msg)
+        existing.password_hash = hashed
+        await session.flush()
```
shoriminimoe: **[10] Redundant `await session.flush()` — `_run_in_session` commits which flushes implicitly.** Extra wire round-trip (UPDATE on flush, then COMMIT) with no behaviour change, and inconsistent with sibling handlers (`user_create`, `user_add_to_tenant`) which rely on commit-time flush. Drop the line.

tests/accounts/test_cli.py
```diff
@@ -0,0 +1,203 @@
...
+    assert second.exit_code != 0
+    assert "already exists" in second.stderr.lower()
+    assert "alice" in second.stderr.lower()
+
+
+def test_user_set_password_succeeds(runner: CliRunner, db_url: str) -> None:
+    runner.invoke(main, ["user", "create", "alice", "--password", "old"])
+
+    result = runner.invoke(main, ["user", "set-password", "alice", "--password", "new"])
+    assert result.exit_code == 0, (result.output, result.stderr)
```
shoriminimoe, line 127: **[11] Test gap: `test_user_set_password_succeeds` asserts exit code 0 only.** Never reads the user back to confirm the new hash persisted. A refactor that turns `existing.password_hash = hashed; await session.flush()` into a no-op (wrong attribute, detached copy) still exits 0 and the test passes — login keeps accepting the OLD password. Follow up with a fresh `UserService.get_by_username('alice')` and `hasher.verify(existing.password_hash, 'new')` assertion.

src/py/novamoc/cli.py
```diff
@@ -0,0 +1,315 @@
...
+    engine = create_async_engine(settings.db.url)
+    try:
+        factory = async_sessionmaker(engine, expire_on_commit=False)
+        async with factory() as session:
+            try:
+                result = await work(session)
+            except BaseException:
+                await session.rollback()
+                raise
+            await session.commit()
```
shoriminimoe: **[12] `await session.commit()` is outside the inner try/except.** A handler whose work-body validation passes but whose commit raises (deferred constraint, future commit-time hook, FK violation once PRAGMA foreign_keys=ON lands per [4]) skips the explicit `session.rollback()` path. `AsyncSession.close()` rolls back implicitly so today there's no data leak, but the code reads as if commit failures are unhandled — a future maintainer adding post-rollback diagnostics or filtered `except` clauses will only cover the work-time branch. Move commit inside the try.

src/py/novamoc/cli.py
```diff
@@ -0,0 +1,315 @@
...
+    "--password",
+    default=None,
+    help="Password for the new user. Prompts interactively when omitted.",
+)
+def user_create(username: str, password: str | None) -> None:
+    """Create a user with a hashed password."""
+    settings = Settings()
+    hasher = _password_hasher(settings)
+    plaintext = _prompt_password_if_missing(password)
+    hashed = hasher.hash(plaintext)
```
shoriminimoe: **[13] `user_create` hashes the password (~100 ms argon2id work) BEFORE the duplicate pre-check.** An idempotent script retrying `novamoc user create $name` pays the full argon2 cost on every retry that hits the 'already exists' branch. Reorder to: open session → pre-check → if free, hash → create. Not a security issue (operator-local CLI), but quality-of-implementation. Same pattern in `user_set_password:200` hashes before the user-not-found check.

src/py/novamoc/cli.py
```diff
@@ -0,0 +1,315 @@
...
+
+    Confirmation is required so a typo at the operator console does
+    not silently set an unrecoverable password.
+    """
+    if password is not None:
+        return password
+    return click.prompt("Password", hide_input=True, confirmation_prompt=True)
+
+
+def _fail(message: str) -> None:
```
shoriminimoe: **[14] `_fail(message)` is dead code.** Defined but never called — every error site uses `raise click.ClickException(msg)` inline. Future contributor reads the helper and is confused by the inline-raise pattern at call sites. Either delete the helper or migrate the inline raises to call it for consistency.

src/py/novamoc/cli.py
```diff
@@ -0,0 +1,315 @@
...
+def tenant_create(display_name: str) -> None:
+    """Create a tenant and print its UUID."""
+    settings = Settings()
+
+    async def work(session: AsyncSession) -> Tenant:
+        svc = TenantService(session=session)
+        return await svc.create(data={"display_name": display_name}, auto_commit=False)
+
+    created = asyncio.run(_run_in_session(settings, work))
+    click.echo(f"Created tenant {created.id}.")
```
shoriminimoe: **[15] `tenant_create` reads `created.id` AFTER `engine.dispose()`.** Works only because `expire_on_commit=False` and `id` is already loaded as a UUID. Fragile contract — a maintainer extending the success line to include `created.display_name` would still work (loaded), but adding any relationship access (`created.memberships`) would crash with `DetachedInstanceError` after the row already committed. Have `work` return a plain dataclass/tuple so the lifecycle contract is explicit.

src/py/novamoc/cli.py
shoriminimoe: fixed in bb06c2c: explicit `TenantService.exists(id=...)` check before the membership insert; phantom UUID now fails with `Tenant '<uuid>' not found.`

src/py/novamoc/cli.py
shoriminimoe: fixed in c399561: `_require_nonblank` rejects empty/whitespace-only on `--password` (both `user create` and `user set-password`) and `--display-name`.

src/py/novamoc/cli.py
shoriminimoe: fixed in a00b34b: narrow `except IntegrityError` added at both UNIQUE backstops (`user_create` and `user_add_to_tenant`); race losers get the same friendly message as the pre-check.

src/py/novamoc/cli.py
shoriminimoe: fixed in 798df1b: `import novamoc.db._listeners  # noqa: F401` added at module top, mirroring `asgi.py`.

src/py/novamoc/cli.py
shoriminimoe: fixed in c643ccc: `_load_settings()` wraps `Settings()` and converts `ValueError` into `ClickException`; every call site routed through it.

tests/accounts/test_cli.py
shoriminimoe: fixed in 6250d41: `test_auth_gc_sessions_deletes_only_expired` seeds one expired and one live `sessions` row, asserts the expired row is gone and the live row remains.

src/py/novamoc/cli.py
shoriminimoe: fixed in e766dd5: `user exists` now exits 0/1/2 (present/absent/CLI-errored). ClickException's exit-1 is intercepted and re-emitted as exit 2 so the absent case stays distinct.

src/py/novamoc/cli.py
shoriminimoe: fixed in c399561: same `_require_nonblank` helper applied to `--display-name` in `tenant create`.

src/py/novamoc/cli.py
shoriminimoe: fixed in 0d04817: each work closure returns the folded `username` (from `UserService.create`'s return value or the pre-checked `existing.username`); the echo lines now print the persisted form.

src/py/novamoc/cli.py
shoriminimoe: fixed in 798df1b: `await session.flush()` dropped from `user_set_password`; `_run_in_session` commits which flushes implicitly.

tests/accounts/test_cli.py
shoriminimoe: fixed in 6250d41: `test_user_set_password_persists_new_hash` re-reads the user row and asserts `hasher.verify(stored, 'new') is True` and `hasher.verify(stored, 'old') is False`.

src/py/novamoc/cli.py
shoriminimoe: fixed in 798df1b: `session.commit()` moved inside the same try as `work(session)`; commit-time failures now follow the explicit rollback path.

src/py/novamoc/cli.py
shoriminimoe: fixed in a00b34b: `get_by_username` pre-check now runs before `hasher.hash()` in both `user_create` and `user_set_password`; retries against an existing/missing name short-circuit before paying the argon2id cost.

src/py/novamoc/cli.py
shoriminimoe: fixed in 798df1b: `_fail` deleted.

src/py/novamoc/cli.py
shoriminimoe: fixed in 798df1b: `tenant_create`'s work closure now returns `uuid.UUID` directly; nothing is read off an ORM instance after `engine.dispose()`.

PR #119, opened 2026-05-25
feat(auth): session-backed Principal+RequestAuth resolver (M5.11, #92)

### LINKED ISSUE #92: M5.11 Replace bearer resolver with session-backed Principal+RequestAuth
The largest single task in the milestone: swap the v1 bearer-token resolver for the session-backed v2 resolver, make `AuthenticationMiddleware` async-with-DB-access, mount the session middleware, and rebuild the test fixture stack. Lands in **one commit** because the resolver shape, the middleware signature, and the conftest fixture all move together — splits leave the test suite red between commits.

## Design pass before implementation

Before writing code, **write a short design spec at `docs/superpowers/specs/<land-date>-auth-middleware-rewrite-design.md`** (the existing repo convention for non-trivial work). Land it in a precursor commit on the same branch; iterate in review if needed. The spec resolves these open implementation-time decisions:

1. **How `AuthenticationMiddleware.authenticate_request` acquires the SQLAlchemy session.** The middleware runs before route handlers, so the controller's `dependencies` map is not yet bound. Options: read `connection.app.plugins.get(SQLAlchemyPlugin).config.get_session()` and open a fresh session for the resolver's duration; reach into the request-scoped session that `before_send_handler="autocommit"` will commit; or other. The first pattern is the cleanest; verify it works against advanced-alchemy's current API.
2. **Where the `PasswordHasher` lives on the app.** `app.state.password_hasher` constructed once at `create_app` time and pulled via a DI provider — confirm this matches the existing `state=State({"settings": s})` pattern in `asgi.py`.
3. **The `dev_admin` test fixture's exact write path.** Direct service calls (`TenantService.repository.add(...)` to pin `id=DEV_TENANT_ID`, then `UserService.create(...)`, then `UserTenantMembershipService.create(...)`) — matching what `just bootstrap-dev` does on the CLI side. The fixture is the only place that pins a specific tenant UUID; the CLI never does.
4. **The `/auth/login` exclusion in the authentication-middleware regex.** Current: `^/(openapi|problems)`. After: `^/(openapi|problems|auth/login)`. Session middleware itself runs for every request including `/auth/login` — login *writes* the session.

The spec should be short (one page is enough); the goal is alignment on the four points above, not a redesign.

## Substantive decisions already made (record in the spec, don't relitigate)

- Resolver shape: `async resolve_principal_from_session(session_payload, users, memberships) -> (Principal, RequestAuth)`. Raises `TenantResolutionError` for: missing session keys; user does not exist; user is disabled; membership for the active tenant is missing.
- Session payload: `{"user_id": "<uuid-str>", "active_tenant_id": "<uuid-str>"}`. Both UUIDs serialised as strings (msgspec / JSON friendly).
- Cookie config: HttpOnly, SameSite=Lax, `Secure` from `AuthSettings.session_cookie_secure`, name from `AuthSettings.session_cookie_name`, `Max-Age` from `AuthSettings.session_ttl_seconds`, `Path="/"`.
- Middleware order in `asgi.create_app`: `[session_config.middleware, DefineMiddleware(AuthenticationMiddleware, exclude=r"^/(openapi|problems|auth/login)"), TenantContextMiddleware()]`.
- `TenantContextMiddleware` is **unchanged** — it still reads `scope["auth"].tenant_id` and calls `use_tenant(...)`. The fact that the type is now `uuid.UUID` instead of `str` (after M5.2) is invisible to it.
- `app.state.password_hasher` is built once at app startup using `AuthSettings` cost params and stashed on `State({"settings": s, "password_hasher": ...})`.
- **No `on_startup` hook for seeding.** The server has no environment-conditional code.

## Code surface

**Modified:**
- `src/py/novamoc/domain/accounts/_resolver.py` — full rewrite (the v1 file becomes the v2 file; the swap-point pattern ADR-017 designed for).
- `src/py/novamoc/domain/accounts/_middleware.py` — `AuthenticationMiddleware.authenticate_request` becomes async with DB access; calls the new resolver.
- `src/py/novamoc/domain/accounts/__init__.py` — drop the `_TENANT_T1_DEV_TOKEN` re-export path; add `AuthController`, `Principal`, `PasswordHasher`, `UserAlreadyHasTenantError` to the public surface.
- `src/py/novamoc/asgi.py` — build the `ServerSideSessionConfig` from `SQLAlchemyAsyncSessionBackend` + `AuthSettings`; mount its `middleware`; extend the authentication-middleware exclude regex; register `AuthController`; register the new problem-details mappers; stash `PasswordHasher` on `app.state`.
- `src/py/novamoc/db/_listeners.py` — small defensive allow-list pin: a module-level `_AUTH_LAYER_TABLE_NAMES = frozenset({"tenants", "users", "user_tenant_memberships", "sessions"})` constant referenced by any listener guard that does column-presence dispatch. The current column-presence heuristic already does the right thing (none of these tables carry `tenant_id`); the pin prevents accidental future regressions.
- `tests/conftest.py` — drop the `_TENANT_T1_DEV_TOKEN` import and the bearer-header default in `client`. Add a `dev_admin` fixture that creates the dev tenant + `admin` user + membership via direct service calls. Rebuild `client` to depend on `dev_admin` and call `POST /auth/login` once on construction. Add an `unauth_client` fixture for the rejection-path tests (no login call).

**Created:**
- `tests/accounts/test_resolver_session.py` — unit tests for `resolve_principal_from_session`:
  - Happy path returns `(Principal(id=..., username="admin"), RequestAuth(tenant_id=DEV_TENANT_ID))`.
  - Missing `user_id` / missing `active_tenant_id` / both missing → `TenantResolutionError`.
  - Unknown user → raises.
  - Disabled user → raises.
  - Active tenant the user has no membership in → raises.

## Acceptance criteria

- The design spec at `docs/superpowers/specs/<land-date>-auth-middleware-rewrite-design.md` is on the branch (precursor commit).
- `uv run pytest` passes — including the existing schema/events e2e tests, which now log in via the rebuilt `client` fixture instead of attaching a bearer header.
- `uv run pytest tests/accounts/test_resolver_session.py` covers all five resolver paths above.
- `uv run ruff check src tests`, `uv run ty check`, `just ratchet` all clean.
- The chosen layering posture from M5.7 (if any exception was carved out) is consistent with this commit.

Exceeds the 8-file heuristic by design; the commit message records the single-conceptual-seam rationale.

**Depends on.** #89 (M5.7 — `Session` model registered), #91 (M5.10 — `AuthController` to mount).

shoriminimoe:
## Summary

Replaces ADR-017's hardcoded-bearer v1 resolver with the session-cookie v2 resolver ADR-020 selected. Lands in two commits on this branch: a precursor design spec, then the swap itself in one commit (the resolver shape, middleware signature, and conftest fixture all move together).

- `resolve_principal_from_session(session_payload, *, users, memberships) -> (Principal, RequestAuth)` folds five failure paths into one `TenantResolutionError` (anti-enumeration per ADR-020).
- `AuthenticationMiddleware.authenticate_request` becomes async-with-DB-access; opens a transient session via `SQLAlchemyPlugin.config[0].get_session()` (the same pattern `SQLAlchemyAsyncSessionBackend` uses internally).
- `asgi.create_app` mounts `SessionMiddleware` + `SQLAlchemyAsyncSessionBackend` ahead of the auth middleware, registers `AuthController`, builds `PasswordHasher` from `AuthSettings`, stashes it on `app.state`, and grows the exclude regex to `^/(openapi|problems|auth/login)`.
- `AuthController` gains a `SecretStr` `type_decoder` (M5.10 left this on the floor — login bodies couldn't decode until now).
- Tests get a `dev_admin` fixture that seeds tenant+user+membership pinned at `DEV_TENANT_ID`; `client` logs in once on construction; `unauth_client` serves rejection-path tests. Three existing `_without_authorization_returns_401` tests rewritten to use `unauth_client` instead of mutating headers.
- PLC0415 ratchet baseline drops 6 → 0 as a side effect of the conftest rewrite eliminating inline imports.

Closes #92. Design spec lives at `docs/superpowers/specs/2026-05-25-auth-middleware-rewrite-design.md` and records the four implementation-time decisions the issue asked to resolve.

Exceeds the 8-file heuristic by design — 16 files in the implementation commit. The spec's "Code surface" section enumerates them; the commit message records the single-conceptual-seam rationale.

## Test plan

- [x] ``uv run pytest`` — 514 passed
- [x] ``uv run pytest tests/accounts/test_resolver_session.py`` — 8 passed (covers all 5 issue-mandated paths plus 3 derived: empty payload, missing tenant key, malformed UUID)
- [x] ``uv run pytest tests/accounts/`` — 95 passed
- [x] ``uv run ruff check src tests`` — only baseline violations
- [x] ``uv run ruff format --check`` — clean
- [x] ``uv run ty check`` — all checks passed
- [x] ``just ratchet`` — baseline matches (Python side; JS coverage ratchet shows the pre-existing "setup error: coverage-summary.json not found" locally because ``npm install`` isn't run in this worktree; CI runs both)


shoriminimoe:
# Code Review

## Overview

Replaces the ADR-017 v1 hardcoded-bearer resolver with the ADR-020 session-cookie design: mounts `SessionMiddleware` + `SQLAlchemyAsyncSessionBackend` ahead of `AuthenticationMiddleware`, makes `authenticate_request` async-with-DB-access (opens a transient session via the plugin's `get_session()`), registers `AuthController`, and rebuilds the conftest fixture stack to log in once rather than attaching a bearer header. 17 files, ~1100/-246. Lands in two commits (design spec, then implementation).

## What's done well

- **Single-seam swap**: the resolver, middleware signature, and conftest all move in one commit — splitting would leave the suite red between commits. Spec calls this out explicitly.
- **Anti-enumeration fold**: 5 failure paths (missing keys, unknown user, disabled user, missing membership, malformed UUID) collapse to one `TenantResolutionError`, matching ADR-020's wire contract.
- **Session-lifecycle parity**: opens the transient session via `alchemy_config.get_session()` — the same pattern `SQLAlchemyAsyncSessionBackend` uses internally — rather than chasing the request-scoped autocommit session that hasn't been bound at middleware time. The `async with` correctly bounds the lifecycle to the call.
- **Real bug catch**: the `SecretStr` `type_decoder` on `AuthController` was an M5.10 leftover that would have made `/auth/login` decode-fail end-to-end. Caught and fixed during integration.
- **Resolver unit tests** (`test_resolver_session.py`) cover all 5 issue-mandated paths plus 3 derived (empty payload, missing tenant key, malformed UUID) — 8 cases, no Litestar app, just the swap-point function.
- **Existing e2e tests** carry forward with mechanical `client`/`unauth_client` edits — the cookie now travels where the bearer header used to.
- **PLC0415 ratchet drop 6 → 0** is a nice side effect of removing inline imports from the old conftest.

## Issues & suggestions

### Spec/implementation drift
The design spec lists `db/_listeners.py` under "Modified (7)" with a `_AUTH_LAYER_TABLE_NAMES` defensive pin, but the implementation commit does **not** touch `_listeners.py`. Either the pin was dropped intentionally and the spec should be updated, or it's a missed item. Worth a quick note in the PR description.

### Resolver: timing side-channel (minor)
`resolve_principal_from_session` issues 0/1/2 SELECTs depending on failure path:
- Malformed UUID → 0 queries (early return)
- Valid UUIDs, unknown user → 1 query
- Valid + known user, missing membership → 2 queries

All produce the same 401, but timing distinguishes them. ADR-020's anti-enumeration claim is about response bytes; timing isn't promised, and network jitter usually swamps it — but recording this as known tech-debt (or running both queries unconditionally) would be a defensible hardening step.

### Middleware vs conftest: duplicated config lookup
`_middleware.py:56-58`:
\`\`\`python
alchemy_config = next(
    c for c in plugin.config if isinstance(c, SQLAlchemyAsyncConfig)
)
\`\`\`
raises bare `StopIteration` on miss (degenerate but ugly). `conftest.py:_async_alchemy_config` does the same thing with a proper `RuntimeError` message. Worth extracting to one helper in `domain.accounts` (or `db/`) and using it in both spots.

### Test fixture concerns
- **Module-level mutable state** in `tests/accounts/test_tenant_context_middleware.py`: `_seen: list[UUID | None] = []` at module scope works because each test calls `_seen.clear()`, but it's brittle under xdist and a fixture would be cleaner.
- **`dev_admin` runs `create_all` manually** before `AsyncTestClient`'s lifespan startup. Idempotent (`CREATE TABLE IF NOT EXISTS`) and the comment acknowledges why, but you could avoid the duplication by having `dev_admin` open its own `AsyncTestClient` (fires the lifespan), then seed inside that scope.
- **`PasswordHasher` built twice** (once in `create_app`, once in `dev_admin`) with identical params. The fixture could pull `app.state.password_hasher` and skip the local construction.

### Exclude-regex anchoring
`exclude=r\"^/(openapi|problems|auth/login)\"` matches `/auth/login/anything` because it's a prefix match. The only handler under `/auth/login` is the exact path today, but a tighter `^/(openapi|problems|auth/login)(/|$)` would prevent accidental future bypass if someone adds `/auth/login/oauth`.

### Principal id typing
`Principal(id=str(user.id), ...)` stringifies the UUID at the resolver boundary, while `RequestAuth.tenant_id` stays a `UUID`. If this is for Litestar's `scope[\"user\"]` historical-string-id convention, fine — but a one-liner comment in `_principal.py` would save the next reader the lookup. Otherwise tightening to `id: UUID` would be more consistent with the rest of the codebase.

## Verdict

Solid, careful work that closes the design loop ADR-017 left open. The spec is rigorous, the implementation matches it (modulo the `_listeners.py` discrepancy), and the test suite proves the contract at the right layer (unit-test the swap-point function, e2e via the cookie-bearing client). Suggest addressing the spec/implementation drift before merge and treating the rest as cleanup follow-ups.

shoriminimoe:
Addressed in `e1a2a61`:

**Spec/impl drift (`_listeners.py`)** — implemented the missed pin. `_AUTH_LAYER_TABLE_NAMES = frozenset({"tenants", "users", "user_tenant_memberships", "sessions"})` lands as a module-level constant in `db/_listeners.py` with the documentation-as-code comment from the spec. Runtime behaviour unchanged (column-presence heuristic already short-circuits these tables); the constant gives a future contributor adding an auth-layer table that *would* carry a `tenant_id` column a greppable place to verify intent.

**Duplicated config lookup** — extracted `pick_async_alchemy_config(plugin: SQLAlchemyPlugin) -> SQLAlchemyAsyncConfig` in `domain/accounts/_middleware.py`. Raises a proper `RuntimeError` on miss (was bare `StopIteration`). Both `AuthenticationMiddleware.authenticate_request` and the test conftest's `dev_admin` import the single helper.

**Exclude regex anchoring** — `^/(openapi|problems|auth/login)` → `^/(openapi|problems|auth/login)(/|$)` with an inline comment explaining the `(/|$)` anchor.

**`PasswordHasher` built twice in `dev_admin`** — pulled `hasher = app.state.password_hasher` instead. The fixture's `settings` parameter dropped since it's no longer used.

**Manual `create_all` in `dev_admin`** — replaced with `await alchemy_config.create_all_metadata(app)`, the same helper the plugin's lifespan startup uses.

**Module-level `_seen` mutable state** — rebuilt as a `probe_seen` fixture returning a fresh `list[UUID | None]` per test; the probe handler closes over it so each test is self-contained. xdist-safe.

**`Principal.id: str` typing** — added a one-line comment on the struct explaining why it stays `str` (avoids stringifying twice per request via `MePrincipal`, which is the wire shape and stays `str` too) while `RequestAuth.tenant_id` is `UUID` (storage-layer ContextVar wants the native type).

**Resolver timing side-channel** — recorded as Tech Debt in the spec rather than changing runtime. v1 ships as-is; the right time to issue both queries unconditionally is alongside login rate-limiting (also a non-shipped concern per ADR-020 §"Consequences").

Also bumped the spec's "Modified" count to 8 to include `controllers/_auth.py` (the `SecretStr` `type_decoder` that landed in the implementation commit).

Verification: `uv run pytest` → 514 passed; ruff/ty clean; ratchet baseline matches.

Issue #13, opened 2026-05-02
Replace placeholder problem-details URN with published docs URL


shoriminimoe:
Per ADR-016, `type` URIs identify failure modes. The current value uses an opaque, non-reachable URN (`urn:novamoc:problems:<code>`) chosen because we have no published documentation hosting yet.

Once we have real docs, replace the URN base with an HTTPS URL whose leaf-segment URLs dereference to per-code documentation (e.g., `https://docs.novamoc.app/problems/name_reserved`). The codes themselves are the stable contract; the URI host changes are non-breaking per RFC 9457 §3.1.

Surfaced in the PR #11 review.

PR #124, opened 2026-05-26
feat(db): bootstrap via Alembic + advanced-alchemy CLI; close M5.15

### LINKED ISSUE #96: M5.15 just bootstrap-dev recipe + README; close #19
Tie the milestone together. Adds the `just bootstrap-dev` recipe, rewrites the `README.md` development-credentials section, runs the final verification matrix, and closes the long-standing #19 tenant-registry tracker.

## `justfile` — new `bootstrap-dev` recipe

Wraps the canonical three-command CLI sequence. Idempotent via the `novamoc user exists` short-circuit (added in M5.13). Suggested shape:

```just
# Create the dev tenant + admin user. Idempotent: skips when admin exists.
# Production runs the equivalent commands in an init container.
bootstrap-dev:
    #!/usr/bin/env bash
    set -euo pipefail
    if uv run novamoc user exists admin >/dev/null 2>&1; then
        echo "admin user already exists; nothing to do."
        exit 0
    fi
    tenant_id=$(uv run novamoc tenant create --display-name "Development" \
                | awk '{print $3}' | tr -d '.')
    echo "Created tenant $tenant_id."
    uv run novamoc user create admin --password admin
    uv run novamoc user add-to-tenant admin "$tenant_id"
    echo "Bootstrap complete. Login at /login with admin / admin."
```

Verify the parse of the `tenant create` output matches what M5.13 actually prints (`Created tenant <uuid>.`); adjust the `awk` field index if needed.

## `README.md` — replace the "Development credentials" section

Replace whatever is currently in the README's auth/credentials section with the new flow. Required content:

- One-liner for local dev: `just bootstrap-dev`.
- Note that production deployments run the equivalent commands in an init container — no environment-conditional code in the server.
- The general CLI shape for adding more tenants/users (`novamoc tenant create --display-name ...`, `novamoc user create <username>`, `novamoc user add-to-tenant <username> <tenant-uuid>`).
- Browser SPA: navigate to `/login`. The auth cookie is HttpOnly and SameSite=Lax; the SPA does not touch the token directly.
- Scripts / curl: `-c cookies.txt` on `POST /auth/login`, then `-b cookies.txt` on subsequent requests.
- One sentence noting that `/openapi` is the only unauthenticated route; everything else returns 401 `tenant_not_resolved` without a valid session.

## Final verification matrix

Run all of the following and confirm clean before merging:

```
uv run pytest
uv run ruff check src tests
uv run ruff format --check src tests
uv run ty check
just ratchet
cd src/js/web && npm run check
```

Then a live-server smoke test mirroring the README's curl example:

```
just bootstrap-dev
uv run litestar --app novamoc.asgi:create_app run --port 8001 &
sleep 2
curl -i -c /tmp/c.txt -X POST http://localhost:8001/auth/login \
    -H 'Content-Type: application/json' \
    -d '{"username":"admin","password":"admin"}'             # expect 204 + Set-Cookie
curl -i -b /tmp/c.txt http://localhost:8001/auth/me           # expect 200
curl -i http://localhost:8001/schema                          # expect 401
curl -i -b /tmp/c.txt http://localhost:8001/schema            # expect 200
kill %1
```

## Closes #19

Commit message ends with `Closes #19.` — the long-standing tenant-registry tracker. The PR body links #19 in the summary so the merge auto-closes.

## Acceptance criteria

- `just bootstrap-dev` runs cleanly on a fresh SQLite (creates the tenant + user + membership; second run prints "already exists").
- The README section is the only documentation surface mentioning the auth flow; no stale references to `_TENANT_T1_DEV_TOKEN` or `Authorization: Bearer t1-dev-token` remain (use `rg` to confirm).
- The full verification matrix above is clean.
- The PR commit message contains `Closes #19.`

**Depends on.** #94 (M5.13 — CLI commands), #95 (M5.14 — `/login` page the README references).

shoriminimoe:
## Summary

Inverts the bootstrap contract: the Litestar app is now a consumer of an already-initialized database. Schema lives behind an Alembic migration tree managed by the advanced-alchemy CLI (`alchemy upgrade head`, `alchemy check`, ...), invoked outside the app's process lifecycle. A fail-fast `on_startup` hook (`db/_startup.assert_alembic_at_head`) refuses to serve when the DB's current revision doesn't match the script tree's HEAD.

Also lands M5.15 (#96): `just bootstrap-dev` recipe + the (previously empty) README's dev-setup section + auth flow docs.

**Closes #96.**

## What's in this PR

**Architecture (ADR-021)**
- `src/py/novamoc/db/config.py` — single chokepoint for `SQLAlchemyAsyncConfig` construction; exposes module-level `alchemy_config` for the alchemy CLI's `--config novamoc.db.config.alchemy_config` flag.
- `src/py/novamoc/db/migrations/` — in-package Alembic tree (env.py + script.py.mako + versions/`03bcd87f986c_baseline.py`) that ships in the wheel via uv_build's module-root.
- `src/py/novamoc/db/_startup.py` — `assert_alembic_at_head()` + `AlembicRevisionMismatchError`. The error message names `just db-init` as the remediation (paste-safe in non-TTY contexts; the recipe already passes `--no-prompt`).
- `src/py/novamoc/asgi.py` — `create_app(settings, *, alchemy_config=None)` test-injection seam; mounts the startup gate.

**Removed**
- `NOVAMOC_DB_CREATE_ALL` env var.
- `DatabaseSettings.create_all` field.
- `create_all=True` plumbing through the plugin and the test `settings` fixture.

**Test path**
- The conftest's `app` fixture builds the engine itself, runs `metadata.create_all` against it, stamps Alembic HEAD via `AlembicCommands(cfg).stamp("head")` (off-threaded — env.py calls `asyncio.run` internally), then hands the populated config to `create_app`. No migrations in the in-memory test path; the gate accepts the stamped DB uniformly.
- `tests/db/test_startup_gate.py` covers both the gate's failure and success paths.

**Operator surface**

```sh
just db-init          # = uv run alchemy --config novamoc.db.config.alchemy_config upgrade head --no-prompt
just db-revision "x"  # generate a new revision
just db-check         # CI gate; joined to `just check`
just bootstrap-dev    # db-init + create Development tenant + admin/admin user (idempotent)
```

**Docs**
- `docs/adr/021-database-bootstrap-via-alembic.md` — four substantive decisions (Alembic via advanced-alchemy CLI, in-package migrations, baseline now, fail-fast gate).
- `CLAUDE.md` — new "Database bootstrap" section; second carve-out under "Critical layering rule" (`db/config.py`'s Litestar import); ADR-021 in pointer list.
- `README.md` — populated from empty; dev-setup + CLI shape + browser-vs-curl auth flow.

## Verification

- 535 tests passing (`uv run pytest`).
- `ruff check`, `ty check`, ruff ratchet all clean on the Python side.
- Live-server smoke: `POST /auth/login` → 204, `GET /auth/me` → 200, `GET /schema` (no cookie) → 401, `GET /schema` (with cookie) → 200.
- Fresh-DB regression: the startup gate fires on a non-initialized DB with `Run: just db-init` in the error.

## Known follow-ups (out of scope for this PR)

- **#126** — `uuid-utils` missing from `[project].dependencies`. Every `alchemy` invocation logs the v4-fallback warning; ADR-020's UUIDv7 commitment isn't being met for `tenants.id` / `users.id` / `sessions.id`.
- **#127** — granian exits `0` when the `on_startup` gate raises. The gate logs the error and stops accepting connections, but a process supervisor relying on exit codes wouldn't detect the misconfig. Likely needs the gate moved from `on_startup` into `create_app` itself; mitigation today is the documented `just db-init`-before-`just serve` discipline.

## Test plan

- [x] Spec written and reviewed
- [x] 12-task implementation plan executed via subagent-driven development (per-task spec + code-quality reviews)
- [x] Final cross-task code review (`approved-for-merge`)
- [x] M5.15 verification matrix (lint / format / typecheck / coverage / ratchet / db-check / live-server smoke / fresh-DB gate)
- [ ] Reviewer signoff


shoriminimoe (commented):
## Code review (extra-high effort, 5 angles + sweep, 15 findings)

Recall-mode review: errs on the side of surfacing. Ranked most-severe first.

---

### 1. `src/py/novamoc/db/_pragmas.py:59` — `foreign_keys=ON` not enabled; FK constraints decorative on SQLite
SQLite ships `foreign_keys=OFF` per-connection. The new pragma listener is the canonical home; this PR sets only `journal_mode=WAL`. Result: every FK declared by `asset_type_fields`, `assets`, `maintenance_records`, `user_tenant_memberships`, etc. is unenforced on SQLite. `cli.py:326-329` already papers over this with a defensive `tenants.exists(id=...)` lookup before creating a membership; other write paths have no such guard and silently accept orphans. Adding `PRAGMA foreign_keys=ON` to `_set_sqlite_pragmas` closes the gap.

### 2. `justfile:59` — `bootstrap-dev` leaks orphan tenant rows on partial failure
Run 1: `tenant create` commits a `Development` tenant; `user create admin` fails (argon2 OOM, validation, Ctrl-C). `set -e` aborts. State: orphan tenant, no user. Run 2: `user exists admin` exits 1; recipe creates ANOTHER `Development` tenant with a fresh UUID and re-runs `user create`. The first tenant lingers forever. Each retry adds another. There's no display-name uniqueness on `tenants`, so the orphans accumulate.

### 3. `justfile:63` — `bootstrap-dev` provisions literal `admin / admin` with no environment gate
`uv run novamoc user create admin --password admin` hardcodes the password. The recipe honors `NOVAMOC_DB_URL` from the shell with no check on the target. An operator running `just bootstrap-dev` against `NOVAMOC_DB_URL=postgresql://prod/...` (misconfigured shell, copy-paste, ambient secrets export) silently provisions `admin/admin` on production. Combined with the weak idempotence check (next finding), the account stays. Gate on URL or require `--allow-prod`.

### 4. `justfile:53` — idempotence check too coarse: `user exists admin` returns 0 even when no membership row was ever written
Run 1 fails at step 5 (`user add-to-tenant`, e.g. broken awk parse — see finding 11). State: tenant + user rows exist, no membership. Run 2 of `just bootstrap-dev`: `user exists admin` exits 0 → recipe prints `admin user already exists; nothing to do.` and exits 0. Login at `/login` succeeds but every authenticated route 401s because the admin has no tenant. Check should be "user-with-membership exists", not "user exists".

### 5. `src/py/novamoc/cli.py:96` — operator CLI bypasses the new Alembic startup gate AND the WAL/(future-FK) pragma listener
CLI commands build their engine via `create_async_engine(settings.db.url)` directly, not via `build_alchemy_config`, and never invoke `assert_alembic_at_head`. Consequences vs. the web app: (1) `novamoc user create admin` against a non-HEAD DB silently writes (or errors with `no such table: users`) instead of the gate's curated `AlembicRevisionMismatchError`; (2) CLI connections never set WAL mode (or `foreign_keys=ON` if added per finding 1). The pre-PR plugin `create_all=True` was the implicit safety net; removing it leaves the CLI more exposed than serve.

### 6. `src/py/novamoc/db/_startup.py:60` — error message names `just db-init`, but production init containers don't have `just`
Production deployment (Docker/Kubernetes init container) hits the gate against an empty/stale DB. The operator reading container logs sees `Run: just db-init` and has no `just` binary; per ADR-021 production runs the upstream `alchemy upgrade head` directly. Suggest either the upstream invocation (`uv run alchemy --config novamoc.db.config.alchemy_config upgrade head --no-prompt`) or a transport-agnostic phrasing ("apply pending Alembic migrations").

### 7. `src/py/novamoc/db/config.py:59` — module-level `alchemy_config = build_alchemy_config(Settings())` parses env vars + materializes an engine at import time
Every importer of `novamoc.db.config` (transitively: `novamoc.asgi`, conftest, alembic env.py, CLI processes that bind `--config`) runs `Settings()` at import. A mistyped env var (`NOVAMOC_AUTH_ARGON2_TIME_COST=oops`) now raises `ValueError` at module load — pytest collection crashes across the whole suite before any test runs, and the CLI's friendly-stderr-on-bad-env contract regresses to a raw ImportError traceback. The eager `cfg.get_engine()` call also creates an `AsyncEngine` pointed at `sqlite+aiosqlite:///novamoc.sqlite` (relative path) that's never disposed. Lazy construction via `functools.cache` or a `get_alchemy_config()` callable closes both holes.

### 8. `src/py/novamoc/db/_startup.py:53` — `version_table_name` is `alembic_versions` (plural) per advanced-alchemy, not Alembic's `alembic_version` (singular)
Verified: `advanced_alchemy/config/common.py:299` sets `version_table_name: str = 'alembic_versions'`. An operator debugging a stuck deployment with raw `alembic current` / `alembic stamp head` (not via the advanced-alchemy wrapper) defaults to the singular name and gets "no current revision" on a DB whose actual table is plural. Misleads them into a destructive `alembic stamp head` or assuming the DB is unmigrated. Either override to Alembic's default or document the discrepancy.

### 9. `src/py/novamoc/db/_startup.py:52` — `get_current_head()` raises `CommandError` on branched migrations
Today the migrations dir has one revision so this never trips. Once a future PR-backport / rebase leaves two parallel heads (common transient state), `assert_alembic_at_head` raises `alembic.util.exc.CommandError('The script directory has multiple heads ...')` instead of the typed `AlembicRevisionMismatchError`. The opaque traceback masks the named `just db-init` remediation. Switch to `get_heads()` with an explicit length-1 check.

### 10. `src/py/novamoc/asgi.py:129` — startup gate doesn't catch generic SQLAlchemy errors
Init container starts before its volume mount completes provisioning; `engine.connect()` raises `OperationalError: unable to open database file`. The gate doesn't catch this; the operator sees a SQLAlchemy stack trace rather than the documented `AlembicRevisionMismatchError` naming the remediation. Same for `database is locked` mid-migration. Wrap the connect in a typed converter that rephrases generic DB errors into the gate's promised UX.

### 11. `justfile:60` — awk parser greedy across all stdout; not anchored to the `Created tenant …` line
`awk '{print $3}'` prints the 3rd field of EVERY stdout line. If `uv run novamoc tenant create` ever emits a preceding stdout line (future tracing config, deprecation notice, a logging hook in a transitive dep), `tenant_id` becomes a multi-line garbage string and `uv run novamoc user add-to-tenant admin "$tenant_id"` fails with "is not a valid UUID". Lines with <3 fields silently inject empty strings. Fix: `awk '/^Created tenant /{print $3; exit}' | tr -d .`.

### 12. `src/py/novamoc/db/migrations/script.py.mako:36` — migration template monkey-patches the global `sqlalchemy` module
Lines 36-47 of the template (and the materialized `03bcd87f986c_baseline.py:42-53`) pollute the global `sqlalchemy` namespace process-wide. Optional `try/except ImportError` blocks assign `typing.Any` (a sentinel) to `sa.Argon2Hasher` etc., so `sa.Argon2Hasher is typing.Any` becomes true. Any third-party library or future test introspecting `sqlalchemy.__dict__` sees these attributes. The baseline already imports `GUID` directly from `advanced_alchemy.types`; the `sa.GUID = GUID` aliasing line is unused by the rest of the file and could be dropped from the template.

### 13. `src/py/novamoc/db/migrations/env.py:90` — non-borrowed-engine branch doesn't register pragmas
In all current paths this branch is unreachable: `AlembicCommands(cfg)` always sets `config.engine` (verified at `advanced_alchemy/alembic/commands.py:385`). But future callers that invoke raw `alembic upgrade head` against the on-disk `alembic.ini` (or any path not going through `AlembicCommands`) hit this branch with no SQLite pragma stack. Migrations apply to a SQLite file that ends in `journal_mode=delete`; first `serve` later flips it to WAL. Defensive: register the pragmas on the newly-built engine too, or document the heuristic.

### 14. `README.md:37` — claims `/openapi` is the only unauthenticated route, but `/problems/*` and `/auth/login` are also excluded
`asgi.py:153` sets `exclude=r"^/(openapi|problems|auth/login)(/|$)"` on `AuthenticationMiddleware`. The README's own `curl -X POST http://localhost:8000/auth/login` example only works because `/auth/login` is unauthenticated, contradicting the very next sentence. Update to: `/openapi`, `/problems/*`, and `/auth/login` are unauthenticated; everything else 401s.

### 15. `src/py/novamoc/db/config.py:56` — `url.startswith('sqlite')` is case-sensitive and substring-only
A URL like `SQLITE+aiosqlite:///foo.db` (uppercase scheme — misconfigured env var, normalisation differences) doesn't match `startswith('sqlite')`, so the WAL pragma listener silently skips registration on a real SQLite engine. Robust alternative: `make_url(settings.db.url).get_backend_name() == 'sqlite'`. Low practical severity but the public-API check survives future drivers / URL formats.

---

shoriminimoe:
Resolution of the 15 review findings, then a force-push to clean up the branch history.

## Resolutions

**Landed in this PR:**

- **#1** — `foreign_keys=ON` added to the pragma listener. Test fallout addressed via two new helper modules (`tests/data/seed_helpers.py` direct-DB seeders + `tests/events/_http_helpers.py` HTTP wrappers). 13 event/snapshot tests migrated to FK-aware seeding. (commit `fix(db): enable PRAGMA foreign_keys=ON`)
- **#3** — bootstrap-dev refuses non-SQLite `NOVAMOC_DB_URL` unless `NOVAMOC_ALLOW_NON_SQLITE_BOOTSTRAP=1`. (commit `fix(just): refuse non-SQLite NOVAMOC_DB_URL`, folded into the bootstrap-dev recipe commit)
- **#5** — CLI's `_run_in_session` now routes through `build_alchemy_config(settings)` and calls `assert_alembic_at_head` before opening a session. `AlembicRevisionMismatchError` maps to `ClickException`. Test fixture stamps HEAD after `create_all` so the gate accepts the seeded schema. (commit `fix(cli): route the operator CLI through build_alchemy_config`)
- **#6** — Startup gate's remediation names both `just db-init` AND the raw `alchemy --config ... upgrade head --no-prompt` so production init containers without `just` get a paste-safe command. (folded into the startup gate commit)
- **#7** — Module-level `alchemy_config` is now lazy via PEP 562 module `__getattr__`. Importing `novamoc.db.config` no longer crashes on a typo'd env var, doesn't materialize an unused `AsyncEngine`. (commit `fix(db): lazy module-level alchemy_config`)
- **#9** — `get_current_head()` replaced with `get_heads()` + length-1 check; branched migrations raise the typed `AlembicRevisionMismatchError` with an `alchemy merge` hint. (folded into the startup gate commit)
- **#10** — `OperationalError` on `engine.connect()` is caught and converted to `AlembicRevisionMismatchError` with the remediation appended. New `test_gate_raises_when_db_unreachable` exercises the path. (folded into the startup gate commit)
- **#11** — awk anchored to `/^Created tenant /{print $3; exit}`. (folded into the small-nits commit)
- **#13** — env.py's non-borrowed-engine branch now calls `register_sqlite_pragmas(connectable)` on the engine it just built. (folded into the small-nits commit)
- **#14** — README accuracy fixed: `/openapi`, `/problems/*`, and `/auth/login` are all unauthenticated. (folded into the small-nits commit)
- **#15** — URL scheme check replaced with `make_url(settings.db.url).get_backend_name() == "sqlite"`. Case-insensitive and robust against `+driver` suffixes. (commit `fix(db): lazy module-level alchemy_config + robust URL scheme check`)

**Deferred to follow-up (#128):**

- **#2** + **#4** — `bootstrap-dev` orphan tenants on partial failure + idempotence check ignoring membership. Proposed remediation is a single `novamoc bootstrap-admin` command that does the whole flow atomically; the recipe collapses to a one-liner. Filed as #128 with the design.

**Declined with rationale:**

- **#8** — Code reads `version_table_name` from `alchemy_config.alembic_config.version_table_name` rather than hardcoding; the docstring's `alembic_versions` reference matches advanced-alchemy's actual default. Not a bug.
- **#12** — The `sa.GUID = GUID` monkey-patching in `script.py.mako` ships from upstream advanced-alchemy. Carrying a local divergence would create maintenance drift; the runtime impact is bounded to module-import-time aliases. Leaving as-vendored.

## History rewrite

Force-pushed: 31 commits → 18. Fixups folded into their parents; review fixes that touched the same surface as their parent were folded in; cross-cutting review fixes are their own commits at the end.

Issue #138, opened 2026-05-30
Epic 1 — Local-first client data engine (ADR-003)


shoriminimoe:
### Problem / motivation

The Python server has shipped its MVP — schema CRUD, event ingestion, snapshot reads, event catch-up, authentication, and tenant scoping are all complete. The Svelte SPA has the schema-management UI and login but no data-side anything: no client-side storage, no asset/record UI. To deliver novaMOC's core promise ("manage maintenance") the frontend needs a local-first data engine, per ADR-003.

Local-first is the highest-risk piece of the frontend roadmap (OPFS support envelope, sync correctness, deterministic LWW fold matching the server's, schema-version gating per ADR-009). The roadmap (`docs/superpowers/specs/2026-05-30-frontend-roadmap.md`) sequenced this as E1 — first, ahead of design system, asset UI, record UI, and WebSocket live sync — to de-risk it before anything builds on top.

### Component

Client (Svelte SPA)

### Proposed solution

Implement ADR-003's local-first storage on the client so subsequent epics read and write through a local SQLite-WASM store synced with the server via the existing HTTP endpoints (`GET /snapshot`, `GET /events`, `POST /events`, `GET /schema`).

Per-epic decisions captured in `docs/superpowers/specs/2026-05-30-epic-1-local-first-engine-design.md`. Headlines:

- **Bundle delivery:** static Vite import of SQLite-WASM.
- **Unsupported browsers:** blocking error UI (no degraded mode).
- **Logout teardown:** delete the OPFS file (not row truncation).
- **Repository layer:** spike issue first (E1.4) → build-out (E1.9).
- **Parity tests:** shared JSON scenarios under `tests/fold-parity/` consumed by pytest and vitest.
- **Schema-change ingest:** refetch `GET /schema` on boot + after each catch-up.
- **Session expiry:** local DB and pending queue survive expiry; re-auth resumes. Explicit logout drops both.
- **Existing schema UI:** stays on HTTP for E1; cutover deferred.

### Acceptance criteria

- [ ] All sub-issues are closed.
- [ ] A fresh-install browser tab syncs a non-trivial tenant snapshot from the server, then creates an event via `recordEvent(...)` that lands in `event_log` on the server.
- [ ] After a server-side schema change, the client's local schema projection refreshes on the next sync.
- [ ] The same fold applied to the same event log on server and client produces identical projection state, verified by the parity-test harness (E1.5).
[...]

Issue #81, opened 2026-05-13
Rewrite Playwright e2e suites for the M4.6 master-detail UI


shoriminimoe:
The three Playwright lifecycle suites under \`src/js/web/tests/e2e/\` (\`asset-type-lifecycle.spec.ts\`, \`field-lifecycle.spec.ts\`, \`maintenance-record-type-lifecycle.spec.ts\`) target the M4.1–M4.5 card-grid UI: \`getByRole('textbox', { name: 'Name' })\`, \`getByRole('button', { name: 'Create', exact: true })\`, etc. The M4.6 redesign (#79) replaces those affordances with a master-detail layout — input is \`aria-label="New type name"\`, submit is "Save", create entry-point is \`+ New ▾\`. The suites were marked \`test.skip\` in the M4.6 PR to keep \`just check\` green.

**Rewrite scope:**

- Walk each suite end-to-end against the new component shape.
- Use the new accessible labels and the new flows (rail click → right pane action row → in-row form swap → confirm bar / dialog).
- Bring the assertions in line with what the manual smoke matrix on #79 covered.
- Drop the \`[pending M4.6 rewrite]\` suffix and the \`test.skip\` markers.

**Refs.** #79 (the M4.6 design pass that obsoleted the M4.1–M4.5 selectors).

PR #196, opened 2026-05-31
feat(client): OPFS feature probe + blocking error UI

### LINKED ISSUE #140: E1.1 — OPFS feature probe + blocking error UI
### Problem / motivation

ADR-003 mandates fail-closed behaviour on browsers without `FileSystemSyncAccessHandle`. A user opening novaMOC in an unsupported browser must get a clear blocking error naming what they need (Chrome 109+, Safari 17+, Firefox best-effort), not a silent in-memory fallback that loses their data on tab close.

Today there is no feature probe and no blocking error path.

### Component

Client (Svelte SPA)

### Proposed solution

At app bootstrap (before any data fetching or routing) run a feature probe:

1. `'storage' in navigator && 'getDirectory' in navigator.storage` — OPFS root accessible.
2. `'FileSystemSyncAccessHandle' in (await navigator.storage.getDirectory()).constructor.prototype` (or equivalent feature test) — sync access handles supported.
3. `crossOriginIsolated === true` — SharedArrayBuffer prerequisite is met (COOP/COEP are already configured in `vite.config.ts` and the server CORS headers).

If any check fails, mount a blocking SvelteKit error page that:

- Names the failed precondition.
- Lists supported browsers/versions explicitly.
- Does NOT proceed to routing or data fetching.

The probe is a small module at `src/js/web/src/lib/db/probe.ts` exporting one async function returning `{ ok: true } | { ok: false, missing: 'opfs' | 'sync_handle' | 'cross_origin_isolation' }`. The layout calls it before initialising any other state.

### Acceptance criteria

- [ ] `src/js/web/src/lib/db/probe.ts` exports an async `probeOpfs()` function with the result shape above.
- [ ] The root layout (`src/routes/+layout.svelte`) calls it before rendering children and routes to the blocking error UI on failure.
- [ ] A vitest test covers each failure branch by mocking the relevant globals.
- [ ] A Playwright test confirms the blocking UI renders when SharedArrayBuffer is unavailable (achieved by serving a route without COOP/COEP for the test only).
- [ ] The error page names supported browsers and the failed precondition.

### Out of scope

- Any attempt at a degraded in-memory mode.
- Detailed browser-version detection beyond what the feature probe gives us (no UA sniffing).
- Localisation of the error message (English only for now).

### Likely affected files

- `src/js/web/src/lib/db/probe.ts` (new).
- `src/js/web/src/routes/+layout.svelte`.
- `src/js/web/src/routes/_unsupported/+page.svelte` (new — or wherever the blocking error renders).
- `src/js/web/tests/component/probe.test.ts` (new).
- `src/js/web/tests/e2e/unsupported.test.ts` (new).

### Alternatives considered

- **Server-side UA sniffing.** Rejected — UA strings are unreliable and the actual capability test is cheap.
- **Lazy probe on first DB access.** Rejected — the user would see a partially-rendered UI before discovering they can't use it.

### Architecture impact

- ADR-003 §"Consequences".

shoriminimoe:
## Summary

Implements **E1.1** (OPFS feature probe + blocking error UI) of Epic 1
(#138 / #140), per ADR-003 §Consequences. The client refuses to run —
with a blocking error naming the missing precondition — on browsers that
can't back SQLite-WASM with an OPFS sync access handle, rather than
silently falling back to an in-memory DB that loses data on tab close.

- `$lib/db/probe.ts::probeOpfs()` returns
  `{ok:true} | {ok:false, missing:'opfs'|'sync_handle'|'cross_origin_isolation'}`.
  It checks, in order: OPFS root reachable (main thread), `crossOriginIsolated`
  (main thread), then whether a worker can actually acquire a sync access
  handle.
- **Sync-handle detection runs in a worker.** `FileSystemSyncAccessHandle`
  is `[Exposed=DedicatedWorker]` — invisible to the main thread — so
  inspecting prototypes there false-negatives in *every* supporting
  browser. `probe.worker.ts` opens OPFS and acquires (then releases) a
  real handle; `probe.sync-handle.ts` drives it with a timeout.
- **COOP/COEP now reach the document.** Vite's `server.headers` apply to
  assets but not the SvelteKit-rendered HTML, so `crossOriginIsolated` was
  always `false` (latent until this probe needed it). A dev-server
  middleware in `vite.config.ts` sets them on every response; the static
  host owns this in prod (ADR-021).
- Root `+layout.svelte` runs the probe before any auth fetch and
  `goto`s `/unsupported?missing=<reason>` on failure; on `/unsupported`
  itself it skips probe + auth to avoid redirect loops.
- `src/routes/unsupported/+page.svelte` names the failed precondition AND
  lists supported browsers (Chrome 109+, Safari 17+, Firefox best-effort).

## Tests

- `probe.test.ts` — orchestration, mocking the worker boundary (no more
  circular global-prototype mocks).
- `probe.sync-handle.test.ts` — the worker driver (ok / fail / error /
  empty-payload / timeout / no-Worker / construction-throws) via a stubbed
  `Worker`.
- `unsupported.test.ts` — the blocking page per reason + fallback.
- `tests/e2e/unsupported.spec.ts` — real-browser failure UI. The
  worker body (real OPFS I/O) is e2e-covered and excluded from jsdom
  coverage.

## Test plan

- [x] `npm run check` — clean (svelte-check + tsc).
- [x] `npm run test` — 72 vitest pass.
- [x] `npm run test:e2e` — runs against the **real API** (no skip flag):
      `api-smoke` passes (the app boots for a supported browser) and the
      `unsupported` specs pass; lifecycle specs stay skipped (#81).
- [x] `just check` — Python suite, lint, format, typecheck, coverage +
      ruff ratchets, db-check all green.
- [x] CI green — all jobs including the `e2e` job (#197 harness).

## Note

Route is `src/routes/unsupported/+page.svelte` (no leading underscore):
SvelteKit treats `_`-prefixed entries under `src/routes/` as private and
omits them from the route tree, so `_unsupported` would be unreachable.

Closes #140

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.coverage-ratchet.json
+++ b/.coverage-ratchet.json
@@ -1,7 +1,7 @@
 {
   "js": {
-    "branch": 50.0,
-    "line": 67.0
+    "branch": 51.0,
+    "line": 68.0
   },
   "python": {
     "branch": 87.0,
--- a/src/js/web/playwright.config.ts
+++ b/src/js/web/playwright.config.ts
@@ -24,6 +24,42 @@ const API_URL = `http://127.0.0.1:${API_PORT}`
 // defined. The cert is self-signed, hence ``ignoreHTTPSErrors`` below.
 const APP_URL = `https://127.0.0.1:${APP_PORT}`
 
+// ``E2E_SKIP_API=1`` runs only the Vite dev server. Some e2e tests
+// (e.g. the OPFS-probe blocking-UI test, issue #140) don't touch the
+// Python API at all — the layout's probe redirects to ``/unsupported``
+// before any backend call — and the existing in-memory API webServer
+// doesn't bootstrap Alembic (ADR-021), so it currently crashes at
+// startup. Until a proper test harness lands (see issue #81), opting
+// out keeps the SPA-only e2e tests runnable.
+const SKIP_API = process.env.E2E_SKIP_API === '1'
+
+const apiWebServer = {
+  command: `uv run litestar --app novamoc.asgi:create_app run --host 127.0.0.1 --port ${API_PORT}`,
+  cwd: '../../..',
+  env: {
+    NOVAMOC_DB_URL: 'sqlite+aiosqlite:///:memory:',
+    NOVAMOC_DB_STATIC_POOL: 'true',
+  },
+  url: `${API_URL}/openapi`,
+  reuseExistingServer: false,
+  timeout: 60_000,
+  stdout: 'pipe' as const,
+  stderr: 'pipe' as const,
+}
+
+const appWebServer = {
+  command: `npm run dev -- --port ${APP_PORT} --strictPort`,
+  env: {
+    NOVAMOC_API_URL: API_URL,
+  },
+  url: APP_URL,
+  ignoreHTTPSErrors: true,
+  reuseExistingServer: false,
+  timeout: 30_000,
+  stdout: 'pipe' as const,
+  stderr: 'pipe' as const,
+}
+
 export default defineConfig({
   testDir: './tests/e2e',
   fullyParallel: false,
@@ -36,32 +72,6 @@ export default defineConfig({
     trace: 'retain-on-failure',
     ignoreHTTPSErrors: true,
   },
-  webServer: [
-    {
-      command: `uv run litestar --app novamoc.asgi:create_app run --host 127.0.0.1 --port ${API_PORT}`,
-      cwd: '../../..',
-      env: {
-        NOVAMOC_DB_URL: 'sqlite+aiosqlite:///:memory:',
-        NOVAMOC_DB_STATIC_POOL: 'true',
-      },
-      url: `${API_URL}/openapi`,
-      reuseExistingServer: false,
-      timeout: 60_000,
-      stdout: 'pipe',
-      stderr: 'pipe',
-    },
-    {
-      command: `npm run dev -- --port ${APP_PORT} --strictPort`,
-      env: {
-        NOVAMOC_API_URL: API_URL,
-      },
-      url: APP_URL,
-      ignoreHTTPSErrors: true,
-      reuseExistingServer: false,
-      timeout: 30_000,
-      stdout: 'pipe',
-      stderr: 'pipe',
-    },
-  ],
+  webServer: SKIP_API ? [appWebServer] : [apiWebServer, appWebServer],
   projects: [{ name: 'chromium', use: { browserName: 'chromium' } }],
 })
--- a/src/js/web/src/lib/db/probe.sync-handle.ts
+++ b/src/js/web/src/lib/db/probe.sync-handle.ts
@@ -1,0 +1,42 @@
+/**
+ * Main-thread driver for the worker-based sync-access-handle probe.
+ *
+ * `FileSystemSyncAccessHandle` is `[Exposed=DedicatedWorker]`, so the main
+ * thread can't detect it by inspecting prototypes — it has to ask a worker
+ * that actually tries to acquire one. See `./probe.worker`.
+ */
+
+const PROBE_TIMEOUT_MS = 5_000
+
+/** Whether a DedicatedWorker can acquire an OPFS sync access handle. */
+export async function checkSyncAccessHandle(): Promise<boolean> {
+  if (typeof Worker === 'undefined') return false
+
+  let worker: Worker
+  try {
+    worker = new Worker(new URL('./probe.worker.ts', import.meta.url), {
+      type: 'module',
+    })
+  } catch {
+    return false
+  }
+
+  try {
+    return await new Promise<boolean>((resolve) => {
+      // Resolve false if the worker never reports back — a hung probe must
+      // not block app boot.
+      const timer = setTimeout(() => resolve(false), PROBE_TIMEOUT_MS)
+      worker.onmessage = (event: MessageEvent<{ ok?: boolean }>) => {
+        clearTimeout(timer)
+        resolve(event.data?.ok === true)
+      }
+      worker.onerror = () => {
+        clearTimeout(timer)
+        resolve(false)
+      }
+      worker.postMessage('probe')
+    })
+  } finally {
+    worker.terminate()
+  }
+}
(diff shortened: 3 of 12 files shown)


shoriminimoe (commented):
## Overview

Implements E1.1 (issue #140, ADR-003): a startup probe that refuses to boot the SPA when the browser can't back SQLite-WASM with an OPFS sync access handle, and a blocking `/unsupported` page that names the missing precondition. Three preconditions checked in cost-order: OPFS root → `crossOriginIsolated` → real worker-acquired sync handle. Also patches a latent dev-server bug where Vite's `server.headers` never reached the SvelteKit-rendered HTML document, so `crossOriginIsolated` was always `false`.

CI is green across all jobs (`lint-py`, `typecheck-py`, `typecheck-js`, `coverage-py`, `coverage-js`, `e2e`, `ratchet`).

## What's done well

- **Cost-ordered checks** — cheap synchronous predicates first, async worker spawn last; first failure short-circuits.
- **Sync-handle detection in a worker** — the only way to truthfully probe `FileSystemSyncAccessHandle` (`[Exposed=DedicatedWorker]`). Main-thread prototype-sniffing would false-negative in every supporting browser. Docstrings explain this clearly.
- **Fail-closed everywhere** — Worker undefined, constructor throws, error event, hung worker (5s timeout), missing payload all collapse to `false`. The matrix is exhaustively covered in `probe.sync-handle.test.ts`.
- **Worker hygiene** — unique probe filename via `crypto.randomUUID()`, sync handle closed, `removeEntry` cleans up, `terminate()` in `finally`.
- **Vite middleware fix** is the right call — `server.headers` doesn't reach the SvelteKit-rendered HTML, and the docstring correctly defers prod to ADR-021 (the static host).
- **`/unsupported` short-circuit in layout** correctly prevents the redirect loop *and* the auth fetch on the blocking page; the `finally` still flips `probeComplete` so the page renders.
- **Coverage discipline** — `probe.worker.ts` is excluded from jsdom coverage (genuinely unreachable there) and explicitly covered by the e2e. Ratchet bumps `js.branch` 50→51 and `js.line` 67→68 — small, justified.
- **Mocking strategy in `probe.test.ts`** uses `vi.hoisted` + `vi.mock` of the worker driver — clean ESM-safe boundary, avoids the previous circular global-prototype mock attempt the PR body mentions.

## Issues

### Bug — literal backticks rendered in user-facing copy

`src/routes/unsupported/+page.svelte` uses double-backtick code spans inside plain string literals:

\`\`\`svelte
detail:
  'novaMOC stores your data locally in OPFS. Your browser does not expose ' +
  '\`\`navigator.storage.getDirectory\`\`, so the app cannot persist anything.',
\`\`\`

In Svelte's text interpolation (`{reason.detail}`), backticks render as literal text — the user sees the backticks around `navigator.storage.getDirectory` verbatim. Affects all three `REASONS` entries (also `FileSystemSyncAccessHandle` and the COOP/COEP header names). Wrap the code references in `<code>` (and split `detail` into segments) — or just drop the backticks if inline code styling isn't worth the markup change.

### Minor

- **`probe.ts` docstring overstates scope**: \"Probe the runtime for SQLite-WASM-over-OPFS support\" — it doesn't actually exercise SQLite-WASM, only the OPFS preconditions. Tighten the wording.
- **`FakeWorker.behavior` is class-level state** in `probe.sync-handle.test.ts`. Vitest runs tests serially within a file by default, so it works, but it's fragile if anyone later flips `--threads` / `fileParallelism`. An instance field set per test would be more defensive.
- **No layout-level integration test** for the probe→redirect path. The e2e covers it end-to-end (and exhaustively), so this isn't required, just slower to catch a regression than a unit-level layout test would be. Fine to skip.

### Nits

- `probe.worker.ts` casts `self as DedicatedWorkerGlobalScope` for `postMessage` but uses bare `self` for `onmessage`. The `/// <reference lib=\"webworker\" />` already widens `self` — the cast is unnecessary.
- The middleware in `vite.config.ts` sets COOP/COEP on *every* response, assets included. Fine for an all-same-origin SPA, but if cross-origin assets enter the picture later they'll need `Cross-Origin-Resource-Policy` to load — worth a comment if/when that changes.

## Risks

- **5-second worst-case boot delay** if the worker hangs. Documented, tested, and acceptable for a fail-closed probe — the alternative (silent in-memory fallback) is the actual hazard ADR-003 forbids.
- **Probe-file leak** if `root.removeEntry` fails after `handle.close()`. Mitigated by the unique `.novamoc-probe-*` prefix — detectable and not a correctness issue. Could add a best-effort sweep on boot if it becomes a real problem.

## Verdict

Approve with one change requested — fix the literal-backticks rendering on `/unsupported` before merging. Everything else is either polish or out of scope.

shoriminimoe:
Review addressed in e8326f2.

**Bug — literal backticks on `/unsupported`** (blocking): fixed. `detail` is now a sequence of text/code `Segment`s rendered with `<code>` (`{#each}` + `{#if seg.code}`) instead of `` ``backtick`` `` spans that Svelte interpolates verbatim. All three reasons updated (`navigator.storage.getDirectory`, `FileSystemSyncAccessHandle`, the two header names). Verified via the Svelte MCP autofixer (clean) and the e2e.

**Minors:**
- `probe.ts` docstring tightened to "Probe the runtime for the OPFS preconditions SQLite-WASM needs" — it doesn't exercise SQLite-WASM.
- `FakeWorker` state: kept external (with a clarifying comment) rather than per-instance — the driver constructs the worker internally, so the test can't inject per-instance config; behavior is set before each call and reset in `beforeEach`. It's safe under `fileParallelism` too, since vitest isolates each file in its own module context (state isn't shared across parallel files); only `test.concurrent` within a file would share it, which this suite doesn't use.
- Layout integration test: skipped per your note — the e2e covers the probe→redirect path end-to-end.

**Nits:**
- Dropped the `DedicatedWorkerGlobalScope` cast in `probe.worker.ts` — you're right, the `webworker` lib reference makes bare `self.postMessage` typecheck (confirmed via `tsc`).
- Added a CORP note to the COOP/COEP middleware for the future cross-origin-assets case.

Local verification: `npm run check` clean, vitest 72 pass, e2e green (api-smoke + unsupported), ruff + coverage ratchets green. CI re-running on the new commit.

PR #203, opened 2026-05-31
feat(client): HLC implementation + persistence (closes #142)

### LINKED ISSUE #142: E1.3 — HLC implementation + persistence
### Problem / motivation

ADR-006 defines a hybrid logical clock that all event-producing participants implement identically. The client needs an HLC implementation that (a) matches the server's format and ordering rules, (b) survives tab reload, and (c) advances correctly on receipt of remote events with HLCs further ahead than local wall time.

### Component

Client (Svelte SPA)

### Proposed solution

Implement an HLC module at `src/js/web/src/lib/db/hlc.ts`:

- HLC string format: `{physical_ms:016}-{logical:05}-{node_id}` (fixed-width, lexicographically sortable, matching ADR-006).
- `now()`: returns a new HLC for a locally-generated event. Three-way max with persisted state, increments logical on collision.
- `receive(remote: HLC)`: advances local state to `max(local, remote, wall)`; returns no value.
- `nodeId` is generated on first launch (`crypto.randomUUID()`-style) and persisted in `sync_state.node_id`.
- The HLC is persisted in `sync_state` after every advance so a reopened tab cannot regress.
- Drift protection: if `wall_time - persisted_physical > 1 minute`, the module emits a `clock_drift_detected` warning (debug surface only); does NOT reject the wall time (the server is the drift authority per ADR-006).

### Acceptance criteria

- [ ] HLC strings produced by the client compare equal-or-greater than HLCs produced earlier within the same tab session.
- [ ] After tab reload, `now()` returns an HLC strictly greater than any HLC issued before reload.
- [ ] `receive(remote)` where `remote.physical > local.physical` advances local state to `remote.physical` (and zeros local logical).
- [ ] `node_id` is stable across reloads (same value persists in `sync_state`).
- [ ] Vitest covers: monotonicity within session, monotonicity across reload, receive-advance, logical-counter cascade on same physical, lexicographic ordering matches numeric ordering.
- [ ] A shared JSON scenario in `tests/fold-parity/` includes a few HLC strings the server's parser accepts.

### Out of scope

- Clock-skew correction beyond what ADR-006 specifies.
- Multi-tab coordination of the HLC (each tab keeps its own; node_id is per-browser-origin not per-tab; safe because logical counter prevents collision within the same physical_ms — same as ADR-006's "no identical HLCs across devices" guarantee applied at tab grain).

### Likely affected files

- `src/js/web/src/lib/db/hlc.ts` (new).
- `src/js/web/tests/component/hlc.test.ts` (new).
- `tests/fold-parity/hlc_basic.json` (new — for parity).

### Alternatives considered

- **Variable-width HLC encoding.** Rejected — ADR-006 pins fixed-width for lexicographic ordering.
- **Server-supplied node IDs.** Rejected — local generation is sufficient and avoids a round-trip on first launch.

### Architecture impact

- ADR-006.

shoriminimoe:
## Summary

- Client-side Hybrid Logical Clock (`src/js/web/src/lib/db/hlc.ts`) per ADR-006: `now()` stamps locally-generated events, `receive(remote)` folds in a remote HLC, both persist before returning. String format `{physical_ms:016}-{logical:05}-{node_id}` matches the server parser so client/server HLCs interleave under plain TEXT ordering.
- `node_id` (generated via `crypto.randomUUID()` on first launch) and the last-issued HLC persist in `sync_state` so a reopened tab can't regress. Drift past one minute surfaces a `clock_drift_detected` `console.warn` only — the server stays the drift authority (ADR-006).
- **Schema change:** added a `sync_state.last_hlc TEXT` column to the client DDL and bumped the migration scaffold to a versioned-step list (`user_version` 1 → 2). Fresh DBs get the column from the `IF NOT EXISTS` DDL; existing DBs get an idempotent `ALTER TABLE ... ADD COLUMN`.
- Cross-language parity: `tests/fold-parity/hlc_basic.json` (client-format HLC strings) + a self-contained `tests/fold-parity/test_hlc_parity.py` asserting the real server parser (`novamoc.domain.events._hlc.HLC.parse`) accepts each. Auto-discovered by pytest; no runner/conftest/justfile wiring.
- Updated the OPFS bootstrap e2e to assert against the `SCHEMA_VERSION` constant rather than a hardcoded `1`.

## Test plan

- [x] `src/js/web/tests/component/hlc.test.ts` (vitest, real `:memory:` DB): monotonicity within session + across simulated reload, receive-advance, logical-counter cascade, lexicographic == numeric ordering, node_id stable across reloads, drift warning.
- [x] `tests/fold-parity/test_hlc_parity.py` green against the real server parser (5 cases).
- [x] `cd src/js/web && npm run check` clean.
- [x] `npm run test` green (95 tests); `npm run test:e2e` green (5 passed, 3 pre-existing API skips).
- [x] `just lint-py` / `typecheck-py` / `typecheck-js` clean; `just coverage` green (620 py + 95 js); ratchet updated (js line 70→72, branch 52→54) and committed.

Closes #142


shoriminimoe (commented):
Review at extra-high recall. 10 findings, most severe first.

### 1. `LOGICAL_MAX` declared but never enforced — client/server parity break

**`src/js/web/src/lib/db/hlc.ts:172`** (and the receive branches at 190/192/194)

Server's `_hlc.py::HLC.now()` raises `OverflowError` when `prev.logical >= LOGICAL_MAX`; client `now()` does `this.#logical += 1` unconditionally. Once logical hits 100000, `String(100000).padStart(5, '0')` returns `'100000'` — 6 chars, no truncation — and format() emits `…-100000-<uuid>`. The server regex `\A(\d{16})-(\d{5})-(.+)\Z` rejects with `HLCParseError`, so every subsequent local event at that physical_ms is unsendable. The constant `LOGICAL_MAX` is defined on line 30 but read nowhere.

### 2. `parse()` has no validation — NaN propagates into persisted state

**`src/js/web/src/lib/db/hlc.ts:99-104`**

A malformed remote in `Hlc.receive()` makes `Number(hlc.slice(0,16))` return `NaN`. `Math.max(local, NaN, wall) === NaN`; every `newPhysical === …` equality is false (NaN !== NaN); execution falls to the else branch with `#physicalMs = NaN`. `#persist` writes `'String(NaN).padStart(16,'0')'` = `'0000000000000NaN-00000-<uuid>'` to `sync_state.last_hlc`. The next reload re-parses NaN, and every stamp this node will ever emit is malformed. Server's `_hlc.py` validates via regex; client should too.

### 3. No test covers the v1→v2 ALTER path

**`src/js/web/tests/component/db-migrations.test.ts:32-55`**

The entire purpose of the STEPS refactor is the v1→v2 ALTER. No test exercises it. On a fresh `:memory:` DB, STEP[0]'s DDL already creates `sync_state` with `last_hlc` (ddl.ts:35), so STEP[1]'s `hasColumn` returns true and the ALTER is skipped — dead code in CI. A real user upgrading from a build at `user_version=1` takes that path for the first time in production. Suggest a test that hand-builds a v1-shape `sync_state` (no `last_hlc`), stamps `PRAGMA user_version=1`, runs `migrate()`, and asserts (1) the column exists, (2) version is bumped, (3) the existing single row survives.

### 4. Fold-parity fixture's max physical_ms exceeds JS `Number.MAX_SAFE_INTEGER`

**`tests/fold-parity/hlc_basic.json:30`**

`Number('9999999999999999') === 10000000000000000` (verified) — well above MAX_SAFE_INTEGER (~9.007×10¹⁵). The fixture's stated purpose is "client-produced HLC strings." Verified: this exact string round-tripped through TS `parse()` + `format()` becomes a 17-digit physical field, breaking the fixed-width invariant. Practical wall times today are ~1.7×10¹² so production isn't at risk near-term, but the parity contract is one-sided here. Either drop the case, switch the client to BigInt for `physical_ms`, or document the protocol's upper bound at MAX_SAFE_INTEGER.

### 5. Drift-detection zero-guard conflates sentinel with legitimate `physical_ms=0`

**`src/js/web/src/lib/db/hlc.ts:204`**

`if (this.#physicalMs !== 0 && wall - this.#physicalMs > DRIFT_WARN_MS)` treats 0 as "no persisted value." Two consequences: (a) the very first `now()` after a fresh open never warns even if the user's wall clock is years off — exactly when drift matters most; (b) if `Hlc.receive()` ever folds in a remote with `physical_ms=0` (the fixture's case 1 is valid), drift detection is suppressed forever. Gate on whether `lastHlc` was null at `open()` instead of comparing `physicalMs` to 0.

### 6. Column ordering of `last_hlc` diverges between fresh DDL and ALTER-upgraded DBs

**`src/js/web/src/lib/db/ddl.ts:194`**

Fresh DBs get `(id, last_seen_seq, active_schema_version, node_id, last_hlc, last_sync_at)`; upgraded DBs get `(…, node_id, last_sync_at, last_hlc)` because SQLite's `ALTER TABLE ADD COLUMN` always appends. Every consumer today uses explicit column lists, so this is silent — but a future `SELECT *` / `rowMode: 'array'` consumer reads positionally and breaks differently across users. Cheapest fix: move `last_hlc TEXT` to the end of the CREATE TABLE so the two shapes agree.

### 7. STEPS append-only docblock contradicts editing DDL in the same PR

**`src/js/web/src/lib/db/migrations.ts:28`**

The docblock says "A step never rewrites an earlier one — a new schema change appends a new step." This PR is itself editing STEP[0]'s body (ddl.ts:35 adds `last_hlc`). The scheme works only because STEP[1] is idempotent via `hasColumn`. A future maintainer reads the rule literally, appends STEP[2] without editing DDL, and a fresh DB never gets the column. Rephrase as: "DDL.ts reflects the latest column set for fresh DBs; STEPS[N>0] patches DBs that predate each new column. Edit DDL and append a STEPS entry together."

### 8. `hasColumn()` interpolates `table` into raw SQL with no whitelist

**`src/js/web/src/lib/db/migrations.ts:54`**

`pragma_table_info('${table}')` is interpolated because the PRAGMA function form rejects bind parameters. The single literal caller today is safe, and the inline comment flags this, but the helper's signature accepts arbitrary `table: string` — a latent footgun. Either inline the pragma at the one call site and drop the helper, or whitelist `table` against `/^[A-Za-z_][A-Za-z0-9_]*$/` before interpolating.

### 9. `Hlc.open()` can resume an old node's clock sequence under a freshly-generated UUID

**`src/js/web/src/lib/db/hlc.ts:136-157`**

Defensive-only today (no production path clears `node_id` while preserving `last_hlc`), but the read order is: read both; if `node_id` missing, generate a new UUID and persist; if `lastHlc` present, adopt its `physicalMs`/`logical`. The returned `Hlc` carries the NEW nodeId with the OLD nodeId's clock state. Subsequent `now()` emits `(oldPhys, oldLogical+1, NEW-uuid)` — monotonic but identity-confused. Either fail loudly when `(node_id null, last_hlc non-null)` is observed, or reset both together.

### 10. Drift threshold exact boundary isn't pinned

**`src/js/web/tests/component/hlc.test.ts:244`**

The test uses 30_000 ms (below) and 61_000 ms (above). Neither pins the `>` vs `>=` semantics. A refactor flipping the comparison silently passes both assertions. Add a case at exactly 60_000 ms asserting no warning (current behavior is strict `>`).

---

**Severity grouping**

- **High**: #1 (overflow), #2 (NaN propagation).
- **Medium**: #3 (missing upgrade test), #4 (fixture vs JS Number).
- **Low**: #5–#10 — latent footguns, doc drift, test rigor.

shoriminimoe:
All 10 findings addressed in 26b54e2 (also merged current `main` in, so the `.coverage-ratchet.json` overlap with #204 is reconciled — combined baseline js line 73 / branch 56).

**High**
1. **Logical overflow** — `now()`/`receive()` now route every logical increment through a `#tick()` that throws at `LOGICAL_MAX`, matching the server's `OverflowError`. No more 6-digit field.
2. **`parse()` NaN** — `parse()` validates against the server's regex (`/^(\d{16})-(\d{5})-(.+)$/`) and throws `HlcParseError`; a malformed remote can't poison persisted state. Test asserts state survives a rejected `receive()`.

**Medium**
3. **v1→v2 ALTER test** — added: hand-builds a v1-shape `sync_state`, stamps `user_version=1`, runs `migrate()`, asserts the column is added, version bumped, and the existing row survives.
4. **Fixture > MAX_SAFE_INTEGER** — capped the max-physical case at `9007199254740991` (MAX_SAFE_INTEGER, still 16 digits, round-trips exactly) and documented the bound in `hlc.ts` + the fixture description. Kept it as `number` rather than pulling in BigInt.

**Low**
5. **Drift zero-guard** — now gated on `#resumed` (whether `last_hlc` was present at open), not `physical_ms == 0`. Added a fresh-clock case + the exact-boundary case.
6. **Column ordering** — `last_hlc` declared last in the CREATE TABLE so fresh and ALTER-upgraded shapes match.
7. **Docblock** — rephrased: edit `ddl.ts` and append a guarded step together.
8. **`hasColumn` interpolation** — whitelists the identifier against `/^[A-Za-z_][A-Za-z0-9_]*$/` before interpolating.
9. **Orphan-clock resume** — `open()` throws when `node_id` is null/empty but `last_hlc` is present, instead of resuming under a fresh UUID.
10. **Drift boundary** — added a case at exactly `60_000` ms asserting no warning (strict `>`); reworked the within-bound test to use a resumed clock so it actually exercises the threshold.

Local: `npm run check` clean, vitest 109, fold-parity pytest 12, e2e 5, full `just coverage` 627 py, ruff + coverage ratchets green.

PR #205, opened 2026-06-02
feat(client): schema-projection ingest + version gating (closes #145)

### LINKED ISSUE #145: E1.6 — Schema-projection ingest + schema-version gating (ADR-009)
### Problem / motivation

Per ADR-009, the client must hold the current schema in a projection and refuse to apply events whose `schema_version > active_schema_version` until the user transitions through the upgrade flow. Without this gate, a server schema change could corrupt the client projection.

Today the client refetches `GET /schema` on every page mount (`src/lib/schema.ts`), but the data is not persisted, not versioned, and there is no gate.

### Component

Client (Svelte SPA)

### Proposed solution

1. Persist the schema projection in the local DB tables added by E1.2 (`asset_types`, `asset_type_fields`, `maintenance_record_types`, `maintenance_record_type_fields`, `schema_change_log`).
2. On boot and after every successful event-catch-up batch, refetch `GET /schema` and reconcile the projection:
   - Read the response's `schema_version`.
   - Upsert all types and fields.
   - Update `sync_state.active_schema_version`.
3. Implement the version gate as a pure helper `gateEvent(event, activeVersion)`:
   - Returns `'apply'` if `event.schema_version <= activeVersion`.
   - Returns `'buffer'` if `event.schema_version > activeVersion`.
4. Buffered events live in a `pending_schema_buffer` table (or `local_pending_events.schema_state = 'buffered'` — implementation chooses). On schema refresh that brings `active_schema_version` up, gated events get re-evaluated.
5. The schema-management UI keeps reading from `GET /schema` directly (per the per-epic decision); the local schema projection is read by the new repo layer for E3 / E4.

### Acceptance criteria

- [ ] `src/js/web/src/lib/sync/schema.ts` exposes `refreshSchema()` (idempotent) and `gateEvent(event)`.
- [ ] After `refreshSchema()`, the local DB's schema tables exactly match the wire response.
- [ ] `sync_state.active_schema_version` advances monotonically.
- [ ] An event with `schema_version > active_schema_version` is buffered, not applied; verified by a vitest test.
- [ ] On the next `refreshSchema()` that bumps `active_schema_version` past the event's version, the buffered event applies.
- [ ] Parity-test scenarios (E1.5) include at least one "buffered then released" sequence.
- [ ] Existing schema-management UI still functions (it continues using `GET /schema`, unchanged).

### Out of scope

- A user-visible "schema changed, please reload" UX (deferred to a future issue; v1 surfaces it on the debug page only).
- Delta-event ingest for schema changes (deferred — refetch is sufficient).
- The schema-upgrade reconciliation flow (events targeting deleted fields). v1 surfaces the buffer count on the debug page; reconciliation UX is a follow-up.

### Likely affected files

- `src/js/web/src/lib/sync/schema.ts` (new).
- `src/js/web/src/lib/db/fold.ts` (extend with `applySchemaProjection`).
- `src/js/web/src/lib/db/types.ts` (extend with schema types).
- `src/js/web/tests/component/schema-ingest.test.ts` (new).
- `tests/fold-parity/schema_version_gating.json` (new scenario).

### Alternatives considered

- **Delta-event ingest.** Out of scope per Q6.
- **No gating; trust ordering.** Rejected — violates ADR-009.

### Architecture impact

- ADR-008 (schema endpoint), ADR-009 (gating), ADR-012 / ADR-019 (projection).
- `docs/superpowers/specs/2026-05-18-schema-changes-catchup-design.md`.

shoriminimoe:
## Summary

- Persist the server-authoritative schema in the local projection and enforce the ADR-009 schema-version gate: an inbound catch-up event whose `schema_version` is ahead of the client's `active_schema_version` is buffered, not applied, until a refetch of `GET /schema` raises the active version past it.
- `src/lib/sync/schema.ts` exposes `refreshSchema()` (idempotent; reconciles the local schema tables to exactly match the wire response, advances `active_schema_version` monotonically, and releases newly-applicable buffered events) and `gateEvent()` (pure `apply`/`buffer` decision).
- `fold.ts` gains `applySchemaProjection`, which flattens the nested `GET /schema` response into the flat local schema tables; `types.ts` gains the schema types.
- Buffered events live in a new `pending_schema_buffer` table (`ddl.ts` + a guarded `2 -> 3` migration step), distinct from the outbound `local_pending_events`.
- A `schema_version_gating.json` fold-parity scenario covers a buffered-then-released sequence; both the pytest and vitest fold-parity runners were extended symmetrically to replay a `gating` block.

## Test plan

- [x] `npm run check` clean (svelte-check + tsc).
- [x] `npm run test` green (120 tests, incl. the new `schema-ingest.test.ts` and the gating parity scenario in the vitest runner).
- [x] `npm run test:e2e` green (OPFS table list updated).
- [x] `uv run pytest tests/fold-parity` green (gating scenario passes in the pytest runner too).
- [x] `just ratchet` clean; coverage baseline bumped (js line 73->74, branch 56->57).

Closes #145


src/js/web/src/lib/sync/schema.ts
```diff
@@ -0,0 +1,286 @@
...
+  projection: SchemaProjection,
+): Promise<void> {
+  // Child tables first: `*_type_fields` FK their parent type.
+  for (const table of [
+    'asset_type_fields',
+    'maintenance_record_type_fields',
+    'asset_types',
+    'maintenance_record_types',
+  ]) {
+    await store.exec(`DELETE FROM ${table} WHERE tenant_id = ?`, [tenantId])
```
shoriminimoe: **FK violation: wholesale `DELETE FROM asset_types` breaks once data exists.**

`assets.type_id` and `maintenance_records.type_id` reference their parent type tables *without* `ON DELETE CASCADE` (see `ddl.ts` — only `asset_type_fields`/`maintenance_record_type_fields` have CASCADE). With `PRAGMA foreign_keys=ON` set in `bootstrap.ts` and `worker.ts`, the very first `refreshSchema` after a single asset has been folded raises `SQLITE_CONSTRAINT_FOREIGNKEY` here, the catch fires `ROLLBACK`, the gate is not advanced, the buffer is not drained, and every subsequent refresh fails the same way — the client cannot recover without wiping the OPFS file.

Tests never fold data before refreshing, so CI passes. This becomes terminal the moment the data fold-to-SQL bridge lands.

Suggested fix: replace the delete-then-insert dance with `INSERT … ON CONFLICT(tenant_id, id) DO UPDATE SET …` (or `INSERT OR REPLACE`) and only delete rows whose ids are absent from the wire response. That preserves the FK chain and is also the right shape for the audit-timestamp issue noted below.

src/js/web/src/lib/sync/schema.ts
```diff
@@ -0,0 +1,286 @@
...
+
+    // Monotonic: never lower the active version, even if a stale read reports
+    // an older one. `MAX` keeps the gate from loosening.
+    await store.exec(
+      `UPDATE sync_state
+          SET active_schema_version = MAX(active_schema_version, ?)
+        WHERE id = 1`,
+      [projection.schema_version],
+    )
+    const activeVersion = await readActiveVersion(store)
```
shoriminimoe: **Stale-snapshot inconsistency: schema rows revert while the gate threshold doesn't.**

`active_schema_version` is bumped via `MAX(active_schema_version, ?)` so it never moves backward — good. But `reconcileTables` runs *unconditionally* and wholesale-replaces the schema rows with whatever the wire returned. If a stale `GET /schema` reports `schema_version=1` after the stored version is already 2, the local schema tables get rolled back to v1's contents while the gate stays at 2.

Now an inbound event tagged `schema_version=2` gates to `'apply'` (sv=2 ≤ 2) and folds, writing a value for a field id (e.g. `COLOR`) that the local schema rows no longer know about — silent projection corruption.

The "monotonic" test (`schema-ingest.test.ts:178`) only asserts `activeVersion === 2`; it never queries the projection tables after the stale refresh, so the regression is uncaught. Either gate the reconcile on `wire.schema_version > stored` (same monotonic check), or fold reconcile + version bump into a single atomic strictly-newer branch.

src/js/web/src/lib/sync/schema.ts
```diff
@@ -0,0 +1,286 @@
...
+       FROM pending_schema_buffer
+      WHERE tenant_id = ? AND schema_version <= ?
+      ORDER BY seq`,
+    [tenantId, activeVersion],
+  )
+  if (rows.length === 0) {
+    return []
+  }
+  await store.exec(
+    'DELETE FROM pending_schema_buffer WHERE tenant_id = ? AND schema_version <= ?',
```
shoriminimoe: **Buffer-to-fold handoff is non-atomic — released events are lost on caller failure.**

The `DELETE` here runs inside `refreshSchema`'s transaction and commits before `refreshSchema` returns the `released[]` to the caller. If the caller crashes or throws between receiving the array and folding the events (OOM, corrupted body JSON in fold, browser tab close, unrelated exception in catch-up loop), those events are:

- Already gone from `pending_schema_buffer`
- Never written to `event_log`
- Below the catch-up cursor on next reload — so the next `GET /events?cursor=` walks past them

The local projection is then permanently missing those mutations until a full snapshot resync. The docstring on `refreshSchema` claims "all run in one transaction so a failed fetch or write leaves the prior state intact" — true *inside* the function's boundary, but the handoff to the caller's fold is the broken seam.

Options: (a) hand the caller a fold callback that runs inside the same transaction; (b) defer the buffer DELETE until the caller acks; (c) write the released events to `event_log` here so a re-fold on next boot recovers them.

src/js/web/src/lib/sync/schema.ts
```diff
@@ -0,0 +1,286 @@
...
+          SET active_schema_version = MAX(active_schema_version, ?)
+        WHERE id = 1`,
+      [projection.schema_version],
+    )
+    const activeVersion = await readActiveVersion(store)
+    const released = await releaseBufferedEvents(store, tenantId, activeVersion)
+    await store.exec('COMMIT')
+    return { activeVersion, released }
+  } catch (error) {
+    await store.exec('ROLLBACK')
```
shoriminimoe: **Nested `BEGIN` breaks composability — and the catch's `ROLLBACK` then aborts the *caller's* outer transaction.**

`refreshSchema` issues raw `BEGIN`/`COMMIT`/`ROLLBACK` via `DbHandle.exec` (separate `postMessage` round-trips on the worker — no per-caller isolation primitive). When an outer caller naturally wraps `{fetch events → release buffered → fold}` in its own `BEGIN` (an obvious shape for the M3 catch-up bridge given the atomicity claim in this module's docstring), the inner `BEGIN` here throws "cannot start a transaction within a transaction." The catch then sends `ROLLBACK` — which targets the only live transaction, i.e. the *outer* one. The outer caller's pending writes silently vanish, and the surfaced error is the nested-BEGIN exception, hiding the data-loss side effect.

`SAVEPOINT`/`RELEASE`/`ROLLBACK TO` is the right primitive when composability is desired. At minimum, the docstring should call out that this function must be called outside any transaction.

src/js/web/src/lib/sync/schema.ts
```diff
@@ -0,0 +1,286 @@
...
+async function upsertTypes(
+  store: SchemaStore,
+  table: string,
+  tenantId: string,
+  types: SchemaType[],
+): Promise<void> {
+  for (const type of types) {
+    await store.exec(
+      `INSERT INTO ${table} (tenant_id, id, name, active) VALUES (?, ?, ?, ?)`,
+      [tenantId, type.id, type.name, type.active ? 1 : 0],
```
shoriminimoe: **Audit timestamps silently nulled on every refresh.**

`asset_types` / `asset_type_fields` / `maintenance_record_types` / `maintenance_record_type_fields` all declare `created_at TEXT, updated_at TEXT` in `ddl.ts`. `upsertTypes`/`upsertFields` only INSERT the structural columns, so the audit columns land NULL on first ingest and stay NULL forever (the delete-then-insert in `reconcileTables` discards any prior value).

The wire `GET /schema` doesn't carry these timestamps, so the local "created_at" was probably meant to mean "first seen locally" — in which case it needs to be preserved across reconciles via the upsert path suggested in the FK-violation comment. Otherwise, drop these columns from the client DDL (matches ADR-008's read-snapshot semantics — the server is authoritative; the client doesn't have a meaningful audit timeline for the schema).

The "idempotent" test selects only the columns it names, so this regression is uncaught.

src/js/web/src/lib/sync/schema.ts
```diff
@@ -0,0 +1,286 @@
...
+    [tenantId, activeVersion],
+  )
+  return rows.map((row) => ({
+    seq: row[0] as number,
+    hlc: row[1] as string,
+    schema_version: row[2] as number,
+    family: row[3] as string,
+    type_id: row[4] as string,
+    instance_id: row[5] as string,
+    body: JSON.parse(row[6] as string),
```
shoriminimoe: **Released events aren't structurally a valid `EventEnvelope` — the type weakening masks downstream regressions.**

`BufferableEvent.body` is `unknown` (because `JSON.parse` returns `unknown`) and `family` is `string` (vs. the fold's `EntityFamily = 'asset' | 'maintenance_record'`). A future caller that writes `fold(projection, result.released)` will hit a TypeScript mismatch and reach for an unchecked `as EventEnvelope[]` — which then lets a buffered body with a typo'd `event` discriminator or an unknown `family` flow into the fold. `_HANDLERS[family]` returns undefined and the fold throws at runtime with no signal that the buffer is the source.

Fix: re-validate at this boundary (cheap discriminated-tag check on `event` + `family`) and return `EventEnvelope[]` so callers don't need to cast.

src/js/web/tests/component/schema-ingest.test.ts
```diff
@@ -0,0 +1,278 @@
...
+      store: db,
+      tenantId: TENANT,
+      client: fakeClient(snapshotV2()),
+    })
+
+    expect(result.activeVersion).toBe(2)
+    expect(await activeSchemaVersion(db)).toBe(2)
+  })
+
+  it('is idempotent — a repeat refresh reproduces the same tables', async () => {
```
shoriminimoe, line 180: **Test gap: monotonicity check doesn't verify the projection rows.**

This test asserts `activeVersion === 2` after a stale v1 refresh, but never re-reads `asset_type_fields` / `maintenance_record_types`. The stale-snapshot bug flagged on `refreshSchema:278` (schema rows revert to v1 while gate stays at 2) passes this test trivially.

Add something like:
```ts
expect(await rows(db, 'SELECT id FROM asset_type_fields WHERE id = ?', [COLOR]))
  .toHaveLength(1)
expect(await rows(db, 'SELECT id FROM maintenance_record_types WHERE id = ?', [INSPECTION]))
  .toHaveLength(1)
```
…and the bug surfaces immediately.

src/js/web/src/lib/sync/schema.ts
```diff
@@ -0,0 +1,286 @@
...
+     VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+    [
+      tenantId,
+      event.seq,
+      event.hlc,
+      event.schema_version,
+      event.family,
+      event.type_id,
+      event.instance_id,
+      JSON.stringify(event.body),
```
shoriminimoe: **`JSON.stringify(undefined)` → `NOT NULL` violation on `body`.**

`BufferableEvent.body: unknown` permits `undefined`. `JSON.stringify(undefined)` returns the JS value `undefined` (not the string `"undefined"`), sqlite-wasm binds it as NULL, and `body TEXT NOT NULL` rejects the INSERT.

In practice, the catch-up code always provides a body — but a `Deactivated`/`Activated` envelope decoded from a sparse server response could land here with `body === undefined`, taking the whole batch down with an unhelpful constraint error. Either tighten the type to `EventBody` (or a discriminated union) so the compiler enforces presence, or pre-check at the boundary and reject with a typed error.

tests/fold-parity/schema_version_gating.json
```diff
@@ -0,0 +1,88 @@
...
+      },
+      {
+        "entity_id": "aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa",
+        "field_id": "33333333-3333-3333-3333-333333333333",
+        "value_json": "red",
+        "hlc": "0000000000000002-00000-client-a"
+      }
+    ],
+    "maintenance_records": [],
+    "maintenance_record_field_values": []
```
shoriminimoe, line 98: **Parity scenario is too lenient to detect a broken `gateEvent`.**

A regression where `gateEvent` always returns `'apply'` would produce the same final projection as the gated path: HLC-ordered LWW makes apply-order irrelevant to the projection state, and both events here eventually become applicable. So the scenario can't distinguish a correctly-gating fold from a never-gating one.

The unit tests in `schema-ingest.test.ts` cover `gateEvent` directly, so this is more of a parity-contract gap than a critical hole. Strengthen by adding an event whose `schema_version` exceeds every phase's `active_schema_version` — its absence from the final projection then becomes the assertion that gating happened.

src/js/web/src/lib/db/fold.ts
```diff
@@ -43,8 +43,81 @@ import type {
...
+ */
+interface WireField {
+  id: string
+  name: string
+  data_type: SchemaField['data_type']
+  validation: Record<string, unknown> | null
+  active: boolean
+}
+
+interface WireType {
```
shoriminimoe: **`SchemaSnapshotWire` duplicates `SchemaSnapshot` (lib/schema.ts) — drift hazard.**

`WireField`/`WireType`/`SchemaSnapshotWire` here and `FieldView`/`TypeView`/`SchemaSnapshot` in `lib/schema.ts` are structurally identical but independently declared — there's no `import`/`extends` linking them. The docstring claims `lib/schema.ts` "re-exports its own structurally-identical view" but it doesn't — they're parallel declarations.

If the server adds a field on `TypeView` (e.g. a `description` for layout/sort), `lib/schema.ts` is updated to receive it but `flattenTypes` here silently drops it. TypeScript accepts the wider input structurally and never flags the mismatch.

Fix: either `interface WireField extends FieldView {}` (and same for the wrappers), or just import `SchemaSnapshot` from `lib/schema.ts` and use it as the wire type. The whole point of the duplication-with-comment was a layering concern, but the comment notes the "back-edge" risk is one-way — `fold.ts` can safely import the wire types without inverting the dependency graph.

shoriminimoe (commented):
Recall-mode review — 10 findings, top three are correctness bugs that block this landing on a non-empty client.

**Headline bugs**

1. **FK violation on `refreshSchema` once data exists** — `reconcileTables` wholesale `DELETE`s from `asset_types`/`maintenance_record_types`, but `assets`/`maintenance_records` reference them *without* `ON DELETE CASCADE`. With `PRAGMA foreign_keys=ON` set, the first refresh after any data event lands raises a FK constraint and the schema can never be re-reconciled. Tests don't catch it because they never fold data before refreshing.

2. **Stale-snapshot data/version inconsistency** — `active_schema_version` is bumped monotonically (`MAX(...)`), but `reconcileTables` runs unconditionally. A stale `GET /schema` reverts the projection tables to vN−k while the gate stays at vN, then accepts events referencing fields the local schema no longer knows about. The monotonicity test only checks the version number, not the rows.

3. **Released events lost on caller failure** — the `DELETE FROM pending_schema_buffer` commits inside `refreshSchema`'s transaction; released events are returned by value. A caller crash between return and fold loses them permanently (gone from buffer, never in `event_log`, below the catch-up cursor on next reload).

**Medium**

4. Nested `BEGIN`/`ROLLBACK` is unsafe under any outer transaction — the catch's `ROLLBACK` then aborts the *caller's* transaction silently.
5. `created_at`/`updated_at` audit columns are silently nulled on every refresh because `upsertTypes`/`upsertFields` don't include them.
6. Released `BufferableEvent.body: unknown` + `family: string` aren't structurally a valid `EventEnvelope` — callers will reach for unchecked casts that mask wire-decode regressions.

**Low / test gaps**

7. Test gap: `'advances the active version monotonically'` doesn't assert table contents — covers up #2.
8. `JSON.stringify(undefined)` → `NOT NULL` violation on `pending_schema_buffer.body`.
9. Parity scenario can't distinguish a correctly-gating fold from an always-applying one (HLC-ordered LWW makes apply-order irrelevant to the final projection).
10. `SchemaSnapshotWire` in `fold.ts` duplicates `SchemaSnapshot` in `lib/schema.ts` with no compile-time link — silent drift hazard.

The gating arithmetic itself (`gateEvent`, the buffer/release ordering) is correct, and the TS↔Python parity runners are symmetric. The issues cluster around the **wholesale reconcile** strategy and the **transaction boundaries** around it — fixing #1 with an `INSERT … ON CONFLICT DO UPDATE` upsert plus #2 with a strictly-newer guard would knock most of the top findings out together.

shoriminimoe:
All 10 addressed in fe5ce9f.

**High**
1. **FK violation** — `reconcileTables` now upserts (`INSERT … ON CONFLICT(tenant_id, id) DO UPDATE`) types + fields, then deletes only ids absent from the wire (`… WHERE id NOT IN (<wire ids>)`). Tombstoned types stay in the wire, so delete-absent never removes a type with dependent assets. New test folds an asset then refreshes — no FK crash, the row survives.
2. **Stale-snapshot revert** — `refreshSchema` reads the stored version first and returns `{ activeVersion: stored, releasable: [] }` without reconciling when `wire.schema_version <= stored`. Reconcile + version write only on a strict increase.
3. **Lost released events** — `refreshSchema` no longer deletes; the field is renamed `released → releasable` (read-only), leaving events in the buffer. New exported `discardBufferedEvents(store, tenantId, seqs)` lets the consumer (E1.8) fold-then-discard atomically; contract documented on both.

**Medium**
4. **Nested BEGIN** — switched to `SAVEPOINT`/`RELEASE`/`ROLLBACK TO … RELEASE`, wrapping only the strict-newer branch, so it composes inside an outer transaction.
5. **Audit timestamps** — upserts set `created_at`/`updated_at` on insert and bump `updated_at` (preserving `created_at`) on conflict; `refreshSchema` supplies the ISO timestamp (fold stays pure). Test asserts `created_at` survives across refreshes.
6. **Invalid envelope** — `BufferableEvent.family: EntityFamily`, `body: EventBody`; validated at both `bufferEvent` and the release read (throws `InvalidBufferableEventError`); releasable returned as fold envelopes, no caller cast.

**Low**
7. Monotonicity test now re-reads the projection tables and asserts no revert.
8. `JSON.stringify(event.body ?? null)` — can't emit `undefined` into the NOT NULL column.
9. Gating scenario gains a seq=3 event tagged `schema_version=3` with a final active version of 2 — it stays buffered forever and its asset is **absent** from `expected_projection`. **Verified the fix bites:** forcing `gateEvent`/`_gate` to always-`apply` made the scenario FAIL in both pytest and vitest; reverted.
10. Single source of truth: the wire type lives in `db/types.ts`; `fold.ts` imports it, `lib/schema.ts` re-exports it. Plain type import between client `lib/` modules — no layering/cycle issue.

Local: `npm run check` clean, vitest 126, fold-parity pytest 13, e2e 5, ratchet green (baseline committed).

PR #207, opened 2026-06-02
feat(client): typed repository layer over the local DB (closes #148)

### LINKED ISSUE #148: E1.9 — Repository layer build-out
### Problem / motivation

E3 (asset UI) and E4 (record UI) will build against the client's local DB. They should NOT write raw SQL — they should use a typed repository façade that hides the storage choice, enforces tenant scoping, and offers methods aligned with the domain (`assetRepo.listByType(typeId)`, `assetRepo.upsert(...)`, etc.).

The shape of the layer is committed in E1.4's spike. This issue implements it.

### Component

Client (Svelte SPA)

### Proposed solution

1. Implement the repository layer using the approach chosen in E1.4.
2. Surface area:
   - `assetRepo` — listByType, getById, upsert, archive, restore, delete.
   - `assetFieldValueRepo` — listByAsset, upsert, clear.
   - `maintenanceRecordRepo` — listByType, getById, upsert, archive, restore, delete.
   - `maintenanceRecordFieldValueRepo` — same shape as assetFieldValueRepo.
   - `schemaProjectionRepo` — read-only views of the schema (consumed by E3/E4 to render type-appropriate widgets).
   - `eventLogRepo` — read-only (used by debug surface and parity tests).
   - `pendingQueueRepo` — listPending, markSent, recordFailure (used by E1.10).
3. Every method goes through a single `withTenant(tenantId)` wrapper so the tenant is always pinned.
4. Tests use a real in-memory SQLite-WASM instance (no mocks — same discipline as the server-side test suite).

### Acceptance criteria

- [ ] All repository methods above are implemented and unit-tested.
- [ ] Tests run against a real SQLite-WASM in-memory DB (the bootstrap supports `:memory:` for tests).
- [ ] Tenant-scoping is enforced — a method called under tenant A cannot see tenant B's rows (verified by a paired test under two tenants).
- [ ] Type signatures forbid passing a row from one tenant into another's `upsert` (compile-time check).
- [ ] No raw SQL in E3 / E4 design follow-up.

### Out of scope

- E3 / E4 callers (those are separate epics).
- Repository over the event log beyond the read access this issue ships.

### Likely affected files

- `src/js/web/src/lib/db/repo/` (new directory).
- `src/js/web/src/lib/db/repo/asset.ts`, `asset_field_value.ts`, `maintenance_record.ts`, `maintenance_record_field_value.ts`, `schema_projection.ts`, `event_log.ts`, `pending_queue.ts` (new).
- `src/js/web/tests/component/repo/` (new).

### Alternatives considered

Captured in E1.4 spike.

### Architecture impact

- ADR-003, ADR-014 (tenant scoping).
- E1.4 spike outcome.

shoriminimoe:
## Summary

- Hand-rolled inline-SQL repository façade over the local SQLite-WASM DB (the approach the E1.4 spike chose; query builders cost too much bundle weight for a local-first client). Lives under `src/js/web/src/lib/db/repo/`.
- Surfaces shipped: `assetRepo` (listByType/getById/upsert/archive/restore/delete), `assetFieldValueRepo` (listByAsset/upsert/clear), `maintenanceRecordRepo` (same shape as assets), `maintenanceRecordFieldValueRepo` (listByRecord/upsert/clear), `schemaProjectionRepo` (read-only type/field views), `eventLogRepo` (read-only listSince/getBySeq), `pendingQueueRepo` (listPending/markSent/recordFailure).
- `withTenant(db, tenantId)` is the single entry point — it pins `tenant_id` into every WHERE/VALUES so a repo bound to tenant A never reads or writes tenant B's rows.

### Compile-time tenant safety

`withTenant<B>(...)` brands the repo set (and every row it returns) with a phantom tenant type `B` (`TenantScoped<T, B> = T & { readonly [BRAND]: B }`, where `BRAND` is a `declare`d unique symbol — zero runtime cost). `upsert` accepts `Writable<T, B> = T & { readonly [BRAND]?: B }`: a freshly-built plain draft (no brand) type-checks for either tenant, but a row read from tenant A's repo carries `[BRAND]: 'a'`, which is **not** assignable to tenant B's `[BRAND]?: 'b'`. So `b.assets.upsert(rowReadFromA)` is a type error:

```
Type '"a"' is not assignable to type '"b"'.
```

`tests/component/repo/tenant_safety.test-d.ts` pins this with `@ts-expect-error` lines that CI's `typecheck-js` job (`svelte-check`) enforces — if the brand ever stops working, an unused-directive error turns the build red. Verified by removing a directive and observing the exact diagnostic above.

## Test plan

- [x] `npm run check` (svelte-check + tsc) — 0 errors, including the `.test-d.ts` brand enforcement.
- [x] `npm run test` — 137 passed (28 new repo tests against a **real in-memory SQLite-WASM** DB, no mocks).
- [x] Paired cross-tenant runtime isolation (`cross_tenant.test.ts`): equivalent rows seeded under two tenant ids in one DB; each tenant's repo sees only its own (list/get/field-values/writes).
- [x] `npm run test:e2e` — 5 passed, 3 skipped (pre-existing API-dependent).
- [x] `just coverage` + `just ratchet-update` — JS line 73→75, branch 56→58; ratchet green. Python untouched.

### Parallel-work note

Stayed inside the E1.9 boundary: only `src/lib/db/repo/**` and `tests/component/repo/**` are added. Did **not** touch `fold.ts`, `sync/`, `ddl.ts`, `migrations.ts`, or `types.ts` — repo-local row types live in `repo/_rows.ts` to avoid colliding with E1.6 (#145). The only shared edit is `.coverage-ratchet.json` (expected; reconcile on second merge).

`pendingQueueRepo.recordFailure` leaves the row queued for retry (the durability guarantee) without mutating it — a persistent status/reason column is E1.10's DDL to add and is out of this issue's boundary.

Closes #148

### DIFF
--- a/.coverage-ratchet.json
+++ b/.coverage-ratchet.json
@@ -1,7 +1,7 @@
 {
   "js": {
-    "branch": 56.0,
-    "line": 73.0
+    "branch": 58.0,
+    "line": 75.0
   },
   "python": {
     "branch": 87.0,
--- a/src/js/web/src/lib/db/repo/_rows.ts
+++ b/src/js/web/src/lib/db/repo/_rows.ts
@@ -1,0 +1,133 @@
+/**
+ * Row shapes the repository layer reads and writes.
+ *
+ * These mirror the client DDL (`ddl.ts`) projection and log tables, modulo the
+ * conventions the repo enforces:
+ *
+ *   - `tenant_id` never appears. The pinned tenant (see `_tenant.ts`) supplies
+ *     it; exposing it on the row type would invite a call site to set it, which
+ *     is exactly the cross-tenant footgun the brand exists to prevent.
+ *   - SQLite affinities are surfaced as their JS-native types: INTEGER booleans
+ *     (`deleted`, `active`) become `boolean`, and `properties` / `value_json`
+ *     TEXT columns become parsed JSON. The mapper in each repo owns the
+ *     (de)serialisation, per the spike (single chokepoint per table).
+ *   - Entity rows carry no `name` column (ADR-015): the name is reconstructed
+ *     from the `col:name` field value, not stored on the projection row.
+ *
+ * Defined here rather than in the shared `db/types.ts` so the repo build-out
+ * does not collide with the parallel fold/ingest work on that file.
+ */
+
+/** A parsed JSON object, as held in `properties` / `value_json` columns. */
+export type Json = unknown
+
+// --- Data projections -------------------------------------------------------
+
+/** An `assets` projection row (sans `tenant_id`). */
+export interface AssetRow {
+  id: string
+  type_id: string
+  properties: Record<string, Json>
+  deleted: boolean
+  row_state_hlc: string
+  created_at: string | null
+  updated_at: string | null
+}
+
+/** What an `assets` upsert writes; audit timestamps are server/fold-owned. */
+export interface AssetDraft {
+  id: string
+  type_id: string
+  properties: Record<string, Json>
+  deleted: boolean
+  row_state_hlc: string
+}
+
+/** A `maintenance_records` projection row (sans `tenant_id`). */
+export interface MaintenanceRecordRow {
+  id: string
+  type_id: string
+  asset_id: string
+  properties: Record<string, Json>
+  deleted: boolean
+  row_state_hlc: string
+  created_at: string | null
+  updated_at: string | null
+}
+
+/** What a `maintenance_records` upsert writes. */
+export interface MaintenanceRecordDraft {
+  id: string
+  type_id: string
+  asset_id: string
+  properties: Record<string, Json>
+  deleted: boolean
+  row_state_hlc: string
+}
+
+/** An `asset_field_values` EAV/LWW row (sans `tenant_id`). */
+export interface AssetFieldValueRow {
+  asset_id: string
+  field_id: string
+  value_json: Json
+  hlc: string
+}
+
+/** A `maintenance_record_field_values` EAV/LWW row (sans `tenant_id`). */
+export interface MaintenanceRecordFieldValueRow {
+  maintenance_record_id: string
+  field_id: string
+  value_json: Json
+  hlc: string
+}
+
+// --- Schema projections (read-only) -----------------------------------------
+
+/** An `asset_types` / `maintenance_record_types` row. */
+export interface TypeRow {
+  id: string
+  name: string
+  active: boolean
+}
+
+/** An `asset_type_fields` / `maintenance_record_type_fields` row. */
+export interface TypeFieldRow {
+  id: string
+  parent_id: string
+  name: string
+  data_type: string
+  validation: Json | null
+  active: boolean
+}
+
+// --- Event log (read-only) --------------------------------------------------
+
+/** An `event_log` row — the local copy of the catch-up stream (ADR-011). */
+export interface EventLogRow {
+  seq: number
+  hlc: string
+  schema_version: number
+  table_name: string
+  type_id: string
+  entity_id: string
+  field_id: string | null
+  op: string
+  value_json: Json | null
+  received_at: string | null
+}
+
+// --- Pending queue (read + status writes) -----------------------------------
+
+/** A `local_pending_events` row — a write generated offline (ADR-013). */
+export interface PendingEventRow {
+  client_seq: number
+  hlc: string
+  schema_version: number
+  table_name: string
+  type_id: string
+  entity_id: string
+  field_id: string | null
+  op: string
+  value_json: Json | null
+  created_at: string | null
+}
--- a/src/js/web/src/lib/db/repo/_sql.ts
+++ b/src/js/web/src/lib/db/repo/_sql.ts
@@ -1,0 +1,39 @@
+/**
+ * Column (de)serialisation helpers shared by the repository mappers.
+ *
+ * SQLite has no boolean or JSON affinity: booleans land as INTEGER 0/1 and JSON
+ * columns as TEXT. Each repo's row mapper is the single chokepoint (per the
+ * spike) that converts between the on-disk representation and the JS-native row
+ * type, so the conversions live here rather than being re-derived per table.
+ */
+
+import type { Json } from './_rows'
+
+/** INTEGER 0/1 (or any SQLite truthy/falsy value) to `boolean`. */
+export function toBool(value: unknown): boolean {
+  return Boolean(value)
+}
+
+/** `boolean` to INTEGER 0/1 for a bind list. */
+export function fromBool(value: boolean): number {
+  return value ? 1 : 0
+}
+
+/**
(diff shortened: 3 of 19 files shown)


shoriminimoe:
# Code Review

## Overview

Adds a hand-rolled inline-SQL repository façade under `src/js/web/src/lib/db/repo/` for the WASM SQLite client. Surfaces seven repos (assets, asset field values, maintenance records, MR field values, schema projection, event log, pending queue) reached only via `withTenant(db, tenantId)`, which pins `tenant_id` into every statement and brands the returned repo set with a phantom tenant type so cross-tenant misuse is a *compile* error. +1638 / −2 across 19 files; 28 new component tests, plus a `.test-d.ts` that pins the brand at the type level.

## Strengths

- **Mirrors server discipline**. Tests use a real in-memory SQLite-WASM DB (no mocks), the cross-tenant suite parallels `tests/schema/test_cross_tenant_isolation.py`, and tenant pinning is structural (passed once via context, threaded into every WHERE/VALUES) the same way the server's listeners are.
- **Phantom brand is well-executed**. `declare const BRAND: unique symbol` + `TenantScoped<T, B>` / `Writable<T, B>` is zero-runtime-cost and the `.test-d.ts` `@ts-expect-error` discipline guarantees CI catches regression. The `B = unknown` default keeps single-tenant call sites ergonomic.
- **Mapper chokepoint** (one `rowTo*` per table, positional `rowMode: 'array'` against an explicit `COLUMNS` constant) is a clean single seam for `INTEGER`↔`boolean` and `TEXT`↔`JSON` conversion.
- **Boundary discipline**: doesn't touch `fold.ts`, `sync/`, `ddl.ts`, or `types.ts`. Repo-local row types live in `repo/_rows.ts` to avoid colliding with E1.6. The only shared edit is `.coverage-ratchet.json`.
- **Docstrings are tight and rationale-heavy** — per repo, the comment explains *why* the contract is shaped the way it is (LWW clear writes JSON `null` instead of `DELETE`, tombstoned rows kept in schema projection, etc.).

## Issues / Suggestions

### 1. `pendingQueueRepo.recordFailure` is misleading (medium)

`src/js/web/src/lib/db/repo/pending_queue.ts`:

```ts
async recordFailure(clientSeq, _reason) {
  // ... We still confirm the row exists and belongs to this tenant so
  // a bad client_seq surfaces.
  await db.exec(
    'SELECT 1 FROM local_pending_events WHERE tenant_id = ? AND client_seq = ?',
    [tenantId, clientSeq],
  )
}
```

The comment claims "a bad `client_seq` surfaces," but the `SELECT 1` result is discarded — nothing surfaces. A bogus `client_seq` is silently swallowed. The `_reason` parameter is also unused. Two clean paths:

- **(preferred)** Either drop the body entirely and leave a `// no-op until E1.10 adds a status column` comment with the durability rationale, **or** actually validate: `if (rows.length === 0) throw new Error('unknown client_seq')`.
- Remove `_reason` from the signature until E1.10 wires it to a column — the test (`recordFailure leaves the row queued for retry`) passes `'network down'` but proves nothing about it.

Right now the API looks like it does something it doesn't, which is the worst of both worlds.

### 2. Unchecked `as Record<string, unknown>` cast on `properties` (low)

`asset.ts` and `maintenance_record.ts`:

```ts
properties: parseJson(row[2]) as Record<string, unknown>,
```

`parseJson` returns `Json` (= `unknown`); the cast trusts the column always contains a JSON object. In practice it always will (the projection invariant), but if the fold ever writes the wrong shape this slips through silently. Not blocking — the assertion matches the invariant the DDL encodes — but worth a one-line `// invariant: assets.properties is always a JSON object (ADR-012)` so future maintainers don't widen it carelessly.

### 3. COLUMNS-to-mapper coupling is positional and unenforced (low)

`COLUMNS` is a string literal and the mapper reads positional indices (`row[0]` … `row[6]`). Reordering the string silently breaks the mapper without a type error. Not worth refactoring for E1.9 — the spike chose inline SQL deliberately — but a future improvement could be a `const COLUMNS = ['id', 'type_id', …] as const` array and `row[COLUMNS.indexOf('id')]` (or a typed tuple decoder). Flag for E1.10/E1.11 retro.

### 4. `Json = unknown` is loose (cosmetic)

`_rows.ts` aliases `Json` to `unknown`, so `value_json: Function` would type-check. A stricter `type Json = null | boolean | number | string | Json[] | { [k: string]: Json }` would catch a wider class of accidental writes at the call site. Trade-off is verbose interop with `JSON.parse`'s `any`. Optional.

### 5. Cross-tenant test premise (cosmetic)

`cross_tenant.test.ts` notes that "the local app opens one tenant's DB file at a time, but the `tenant_id` column is carried on every row for wire/fold parity." Worth being explicit that **this test deliberately co-locates two tenants in one DB to prove the WHERE/VALUES pinning** — that's the property E3/E4 actually rely on. Fine as-is; just calling out the framing is intentional.

## Correctness, Conventions, Performance, Security, Tests

- **Correctness**: All statements are parameterised — no string interpolation of user data. Tenant ID flows from context, not from arguments to individual methods.
- **Conventions**: Underscore-prefixed `_*.ts` for internal modules matches the no-collision strategy described in the PR. File naming follows table names (snake_case). Docstrings are Napoleon-equivalent concise.
- **Performance**: No N+1 risks in the repo surface; `listByAsset` / `listByRecord` returns one row per field per asset (the EAV shape). `event_log` uses `LIMIT` with a default of 500.
- **Security**: Parameter binding throughout; the brand makes cross-tenant misuse a compile error, and the runtime contract is *also* enforced by the SQL itself (every WHERE includes `tenant_id = ?`). Belt-and-braces, the right level for client-side data.
- **Tests**: 28 new tests, real in-memory DB, paired cross-tenant runtime suite + `.test-d.ts` for the type-level guarantee. Coverage ratchet correctly bumped (73→75 line, 56→58 branch).

## Verdict

**Approve with a small ask on `recordFailure`** — please either make it actually validate `client_seq` (and remove the now-meaningless `_reason` until E1.10 has a column for it) or drop the SELECT and let the comment stand. The rest is solid; the phantom-brand work in particular is a nice piece of leverage that future repos (and reviewers) get for free.

shoriminimoe:
Addressed in 8ec84a4.

**#1 `recordFailure` (the ask):** now throws on an unknown `client_seq` instead of discarding the `SELECT` — so draining a row that isn't there surfaces loudly. Dropped the `reason` param (E1.10 re-adds it alongside the status/reason column it needs). Updated the test to drop the meaningless `'network down'` arg and added a case asserting the unknown-`client_seq` throw.

**#2 `properties` cast:** added the `// invariant: properties is always a JSON object (ADR-012 fold)` note at both mapper casts (`asset.ts`, `maintenance_record.ts`).

**#3/#4/#5** (positional COLUMNS coupling, `Json = unknown`, cross-tenant test framing): left as-is per your "flag for later / optional / cosmetic" — happy to revisit the COLUMNS→mapper coupling in an E1.10/E1.11 retro.

`npm run check` clean, vitest 138, coverage ratchet regenerated + green.

Issue #38, opened 2026-05-03
M3.3 Fan-out hook on event acceptance


shoriminimoe:
After the M1 accept path commits a batch, publish the accepted events to the subscriber registry. Publish is post-commit (a rolled-back transaction must not fan out). Wire as a dependency-injected callback on the events service so tests can stub it.

**Refs.** ADR-011, ADR-013.
**Depends on.** #24 (M1.5 — accept path), #37 (M3.2 — registry).

PR #209, opened 2026-06-05
feat(client): snapshot ingest from GET /snapshot

### LINKED ISSUE #146: E1.7 — Snapshot ingest from GET /snapshot
### Problem / motivation

On first login (or after explicit "drop local DB and resync"), the client needs to hydrate from the server's bulk snapshot endpoint before incremental catch-up can begin. ADR-013 / ADR-015 define the snapshot envelope and the page-token pagination.

### Component

Client (Svelte SPA)

### Proposed solution

1. New module at `src/js/web/src/lib/sync/snapshot.ts`.
2. Implements: `ingestSnapshot(): Promise<{ cursor: number, schema_version: number }>`.
3. Loop:
   - Fetch `GET /snapshot?page=<page>` (initially no page).
   - Decode the batch via the discriminated-union body (`assets`, `asset_field_values`, `maintenance_records`, `maintenance_record_field_values`).
   - Apply each row to the local projection using the same fold rules from E1.5 (treating snapshot rows as committed projection state, not events — the `*_field_values` rows already carry HLC).
   - On a non-terminal batch, recurse with the returned `page` token.
   - On the terminal batch, persist `cursor` → `sync_state.last_seen_seq` and `schema_version` → `sync_state.active_schema_version`. Return both.
4. Blocking-modal-with-progress UI on the debug surface (`/_debug/sync`) shows per-table progress (rows seen / total batches).
5. Failure semantics: on transport error mid-pagination, the partial state stays in the DB and the next call resumes from the persisted `page` token — but only if the server's `start_seq` (encoded in the token) is still valid. If the server has advanced past it, the snapshot restarts.

### Acceptance criteria

- [ ] `ingestSnapshot()` returns the terminal `(cursor, schema_version)`.
- [ ] After ingest, all four projection tables contain the same rows the server would return via direct query (verified by a Playwright integration test against a seeded tenant).
- [ ] A network failure mid-snapshot resumes correctly from the persisted `page` token on the next attempt.
- [ ] An invalidated page token (server advanced past `start_seq`) is detected and restarts the snapshot from scratch.
- [ ] The blocking modal shows progress per table.
- [ ] Parity-test scenario added: snapshot-then-catch-up equivalence.

### Out of scope

- Historical (as-of-seq) snapshots — not supported by the server.
- Snapshot progress UI in the main app shell (debug-only for E1).
- Performance tuning for snapshots > 10k rows.

### Likely affected files

- `src/js/web/src/lib/sync/snapshot.ts` (new).
- `src/js/web/src/lib/sync/_progress.ts` (new — small progress observable).
- `src/js/web/src/lib/db/fold.ts` (extend with `applySnapshotRow`).
- `src/js/web/src/routes/_debug/sync/+page.svelte` (progress UI lives here; full debug surface is E1.12).
- `src/js/web/tests/component/snapshot-ingest.test.ts` (new).
- `src/js/web/tests/e2e/snapshot-ingest.test.ts` (new).
- `tests/fold-parity/snapshot_round_trip.json` (new scenario).

### Alternatives considered

- **Single-shot snapshot, no pagination.** Server doesn't support it; rejected.
- **Resume by re-querying without the page token.** Possible but wastes bandwidth.

### Architecture impact

- ADR-013, ADR-015 (snapshot envelope).
- `docs/superpowers/specs/2026-05-18-snapshot-endpoint-design.md`.

shoriminimoe:
## Summary

- Add `src/lib/sync/snapshot.ts::ingestSnapshot`, the bulk-snapshot ingest (ADR-013 / ADR-015): walks the paginated `GET /snapshot` transfer, applies each committed row to the four local projection tables, and persists the terminal replication `cursor` -> `sync_state.last_seen_seq` and `schema_version` -> `sync_state.active_schema_version`.
- Extend `db/fold.ts` with `applySnapshotRow` (pure, in-memory) honouring the fold's materialization contract: structural entity columns + field-value tables with `hlc` preserved, no `name`/`properties`.
- Resumable failure semantics: a transport error mid-pagination leaves partial state and a checkpointed `page` token in `sync_state` (new `snapshot_page` / `snapshot_schema_version` columns, v3->v4 guarded migration), so the next call resumes; a batch reporting a different `schema_version` means the server advanced past the snapshot, so the ingest discards partial state and restarts from scratch.
- Add `_progress.ts` (per-table progress observable) and a minimal blocking-modal debug surface at `/_debug/sync` (full surface is E1.12).
- Tests: vitest component (full ingest, resume, invalidation), Playwright integration (seed via real API, ingest in-browser, assert table parity), and a `snapshot_round_trip` fold-parity scenario wired symmetrically into both runners (snapshot-rows-fold == event-fold, independent of catch-up E1.8).

Closes #146

## Test plan

- [x] `cd src/js/web && npm run check` clean
- [x] `npm run test` green (173 tests)
- [x] `npm run test:e2e` green incl. `snapshot-ingest.spec.ts`
- [x] `just check` CI-relevant jobs green: lint-py (ratchet), typecheck-py, typecheck-js, coverage-py, coverage-js, e2e, ratchet (coverage baseline updated + committed)
- [ ] `db-check` not run locally (CI doesn't run it)


shoriminimoe:
### Code review

Found 1 issue:

1. `ingestSnapshot` invalidation branch wipes the local projection and the in-memory `page`/`snapshotVersion` but never calls `persistInFlight({page: null, schema_version: null})` before `continue`. If the process is interrupted between `clearProjection` and the next batch's `persistInFlight`, `sync_state.snapshot_page` and `sync_state.snapshot_schema_version` still hold the stale (pre-invalidation) checkpoint while the projection tables are empty. The next `ingestSnapshot()` call reads a non-null `page`, skips the upfront `clearProjection`, and attempts to resume from the invalidated token — contradicting the module docstring's "restart from scratch" guarantee and the `ddl.ts` invariant that both columns "are `NULL` when no ingest is in flight."

   https://github.com/shoriminimoe/novamoc/blob/71d9091a5c2d8485f1d4531aca4d36c15681be21/src/js/web/src/lib/sync/snapshot.ts#L281-L291

   Invariant from `ddl.ts`:

   https://github.com/shoriminimoe/novamoc/blob/71d9091a5c2d8485f1d4531aca4d36c15681be21/src/js/web/src/lib/db/ddl.ts#L214-L221

   Fix: `await persistInFlight(store, { page: null, schema_version: null })` immediately after `clearProjection` in the invalidation branch (and ideally a vitest case exercising a reload between `clearProjection` and the first batch of the restart).

shoriminimoe:
Fixed in a4bddba.

The invalidation branch now nulls the checkpoint **before** wiping the projection: `persistInFlight(store, { page: null, schema_version: null })` then `clearProjection(...)`. I ordered it null-first (rather than after `clearProjection` as suggested) because it closes the window completely — if the process dies between the two writes, `snapshot_page` is already NULL, so the next run hits the upfront `clearProjection` (page === null) and restarts from scratch rather than resuming the invalid token. Either order satisfies the `ddl.ts` invariant, but null-first leaves no interruptible gap.

Added the vitest case you asked for: batch 1 (v7) checkpoints PAGE-1, batch 2 reports v8 (invalidation), then the restart's first fetch throws — simulating a crash in exactly that window. It asserts `snapshot_page`/`snapshot_schema_version` are NULL and the projection is empty, and that a subsequent "reload" run restarts from `/snapshot` (not the invalidated PAGE-1).

Also brought the branch up to date with `main` (it was 13 commits behind — the server-side broadcaster work, #38/#208): clean merge, combined coverage baseline verified (js 79/61, py 97/88), 653 pytest pass, ratchet green.

